PR #5, opened 2020-10-22
remove envwatcher

tychoish:
cemanater instances will take responsibility for the envwatcher functions.


f0ssel:
I think we need to set the `AUTOOFF_ENABLED` env var in the cemanager deployment spec.

f0ssel (approved):

PR #7, opened 2020-10-26
propogate autooff_enabled to cemanager

tychoish:


deansheather (changes requested):
This needs to be added to `values.yaml` too.

wbobeirne (approved):
Should we add a (real) deprecation notice here for 1.13.0?

deansheather (approved):

tychoish:
I'm going to close this because I don't think it makes sense to given that the envvar is no longer read by the application.

PR #10, opened 2020-12-03
Add cemanager url and creds env vars for envproxy detachment

f0ssel:
This enables communication between envproxy and cemanager, which will soon be the only way for envproxy to handle authenticating requests. Right now we generate a shared user and key secret between both services, this will probably change in the future but works for now.

### DIFF
--- a/templates/_common.tpl
+++ b/templates/_common.tpl
@@ -98,3 +98,12 @@ tolerations:
 {{ toYaml .Values.serviceTolerations }}
 {{- end }}
 {{- end }}
+{{/*
+  coder.envproxyCredsEnv - temporary POC for intra-service auth
+*/}}
+{{- define "coder.envproxyCredsEnv" }}
+- name: "CODER_SERVER_USER"
+  value: {{ now | date "2006-01-02" | sha256sum }}
+- name: "CODER_SERVER_KEY"
+  value: {{ now | date "2006-01-02" | sha256sum }}
+{{- end }}
--- a/templates/envproxy.yaml
+++ b/templates/envproxy.yaml
@@ -69,6 +69,9 @@ spec:
 {{- include "coder.environments.configMapEnv" . | indent 12 }}
 {{- include "coder.devurls.hostEnv" . | indent 12 }}
 {{- include "coder.postgres.env" . | indent 12 }}
+{{- include "coder.envproxyCredsEnv" . | indent 12 }}
+            - name: CEMANAGER_ACCESS_URL
+              value: "http://cemanager:8080"
           command:
             # Bash needs to be used for signals to be passed down correctly.
             - /bin/bash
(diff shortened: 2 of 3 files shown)


templates/envproxy.yaml
```diff
@@ -69,6 +69,9 @@ spec:
 {{- include "coder.environments.configMapEnv" . | indent 12 }}
 {{- include "coder.devurls.hostEnv" . | indent 12 }}
 {{- include "coder.postgres.env" . | indent 12 }}
+{{- include "coder.envproxyCredsEnv" . | indent 12 }}
+            - name: CEMANAGER_ACCESS_URL
+              value: "http://cemanager:8080"
```
f0ssel, line 74: This only works inside the same cluster, but will always work with the default envproxy we ship.

templates/_common.tpl
```diff
@@ -98,3 +98,12 @@ tolerations:
...
 {{- end }}
 {{- end }}
+{{/*
+  coder.envproxyCredsEnv - temporary POC for intra-service auth
+*/}}
+{{- define "coder.envproxyCredsEnv" }}
+- name: "CODER_SERVER_USER"
+  value: {{ now | date "2006-01-02" | sha256sum }}
+- name: "CODER_SERVER_KEY"
+  value: {{ now | date "2006-01-02" | sha256sum }}
```
deansheather, line 108: CODER_SERVER sounds too much like CODE_SERVER imo, so maybe we should call it something else.

templates/envproxy.yaml
deansheather: Doesn't this need the cluster domain suffix appended to it?

templates/_common.tpl
sreya: what about just `SERVICE_KEY`

templates/envproxy.yaml
f0ssel: it does not, if you are in the same namespace this resolves correctly from my testing.

templates/_common.tpl
f0ssel: Yeah this is obv not ideal but I just need a shared secret, and don't want to force users to provide one because it doesn't really matter. This is most likely going to be different before release, it's just to tie things together while we develop this feature.

templates/envproxy.yaml
deansheather: If you're confident that it'll work for every customer than I'm OK with it. For postgres though we have been using the full domain: https://github.com/cdr/enterprise-helm/blob/849b0697001da716a505ee8910671145d87a8213/templates/_common.tpl#L27

f0ssel: Yeah I'll add it to be safe then

f0ssel:
Confirmed the internal service URL still works on https://github.com/cdr/c/pull/6063

deansheather (approved):

PR #17, opened 2021-01-08
Create envproxy-only chart

deansheather:
# !!! DO NOT MERGE THIS PR !!!

This PR exists only for comments/review and diff. It should never be merged as it's basically a fork.

This is what customers will use (via coder-cli) in 1.15 to deploy a non-local alpha workspace provider to another K8s cluster. This will be tidied and replaced by a neat subchart when workspace providers hits beta.

### Changes

- Removes cemanager, dashboard, timescale
- Removes unnecessary templates for RBAC stuff
- Removes unnecessary values
- Removes cemanager/dashboard from ingress rules
- Removes `CODER_SERVER_USER` and `CODER_SERVER_KEY` env vars from envproxy
- Important envproxy vars:
    - `ENVPROXY_ACCESS_URL`:  URL to report to cemanager, taken from `envproxy.accessURL` or derived from `ingress`
    - `SSH_ENABLED`: `true` or `false` to report to cemanager
    - `CEMANAGER_ACCESS_URL`: the full URL used to connect to cemanager, minus `/api`
    - `CEMANAGER_TOKEN`: the envproxy's token
    - `NAMESPACE_WHITELIST`: the namespace whitelist, comma separated (to report to cemanager)
    - `DEVURL_DOMAIN`: the dev URL domain (also reported to cemanager)

Devurls TLS can be sniffed from the envproxy access URL (assume TLS is enabled if envproxy has TLS).


templates/_common.tpl
```diff
@@ -99,11 +56,19 @@ tolerations:
...
+    {{- if .Values.ingress.tls.enable -}}
+    https://
+    {{- else -}}
+    http://
+    {{- end -}}
+    {{- .Values.ingress.host }}/ingress
+{{- else }}
+{{- fail "envproxy.access.URL or ingress.host must be set" }}
+{{- end }}
+{{- end }}
```
f0ssel, line 74: Does this implicitly not support raw IPs?

f0ssel: If so, I think that's OK for alpha? But we will need to in the future, may have to poll kube for it on the ingress from envproxy itself or something on start.

templates/envproxy.yaml
```diff
@@ -128,11 +119,94 @@ spec:
...
+{{ $namespaces := append .Values.namespaceWhitelist .Release.Namespace }}
+{{- range $namespaces }}
+---
+# The envproxy role lists the permissions the envproxy requires.
+# It conditionally sets the pod security policy based on the
+# helm value.
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: envproxy
```
f0ssel, line 140: shouldn't need this anymore

values.yaml
```diff
@@ -68,92 +65,16 @@ devurls:
...
-      cpu: "250m"
-      memory: "512Mi"
-
 # envproxy contains configuration for the service handling long-lived
 # connections to environments such as IDE or shell sessions.
 envproxy:
+  # envproxy.accessURL -- The URL reported to cemanager. Must be accessible by
+  # cemanager and all users who can use this workspace provider. Derived from
+  # ingress.host if not set. This should be a full URL, complete with protocol
+  # and trailing "/ingress" (no trailing slash).
```
f0ssel, line 76: what does `/ingress` mean here?

values.yaml
```diff
@@ -182,55 +103,14 @@ envproxy:
...
-  # e.g.
-  # tolerations:
-  # - key: "my key"
-  #   operator: "Exists"
-  #   effect: "NoExecute"
-  #   tolerationSeconds: 6000
-  tolerations: []
+# cemanager -- Details for connecting to the cemanager instance.
+# NOTE: the envproxy must be accessible by envproxy.accessURL from the
+# cemanager.
```
f0ssel, line 108: should this note be on the envproxy accessURL instead? And here we say "cemanager must be accessible"

templates/_common.tpl
deansheather: Yeah to support the raw IP of the load balancer you'd have to do it like that. Alternatively, the user could deploy with a fake envproxy access URL, wait for load balancer IP, and redeploy with real IP.

values.yaml
deansheather: oops I meant `/proxy`

deansheather:
Closing in favor of #20

PR #18, opened 2021-01-08
Rework envproxy and cemanager environment variables for RPs

deansheather:
Reworks env vars to match the sister PR #17. That contains a fuller list of changes.


templates/cemanager.yaml
```diff
@@ -37,9 +37,9 @@ spec:
       # terminationGracePeriodSeconds should be set to the upper bound for container rebuilds and creates.
       # 5 minutes
       terminationGracePeriodSeconds: 300
-      # Start the manager with a specific service account
-      # that only has permissions to what it needs in order to
-      # administer pods.
+      # Start the manager with a specific service account with no permissions.
+      # The envproxy forwards a service account token to the cemanager on
+      # startup called "coder" with all the permissions it needs.
```
f0ssel, line 42: So envproxy is now the sole owner of the service account (that does anything), as in that service account will be the separate envproxy helm chart?

templates/envproxy.yaml
```diff
@@ -28,7 +28,9 @@ spec:
       restartPolicy: Always
       # 4 hours
       terminationGracePeriodSeconds: {{ .Values.envproxy.terminationGracePeriodSeconds }}
-      serviceAccountName: envproxy
+      # envproxy itself doesn't use this service account, but instead forwards
+      # the account token to the cemanager.
+      serviceAccountName: coder
```
f0ssel, line 33: This means envproxy itself can also easily make kube calls. There could be a future where envproxy runs the controllers and sends build events back to cemanager. (this idea was in the OG resource pools rfc I think) to keep the kube data local and move the responsibility out of cemanager altogether.

f0ssel (approved):
LGTM, When does this get merged in relation to the others stuff?

templates/envproxy.yaml
deansheather: I had to do this because it's difficult to mount a second service account unless you know the secret's name exactly AFAIK.

templates/cemanager.yaml
deansheather: This will be the same in the envproxy chart. Envproxy will assume the `coder` SA, and forward that to the cemanager.

deansheather:
I'll test this in my branch by checking out this submodule branch in `m`, after you merge your branch into master in `m` and I rebase on that.

templates/envproxy.yaml
f0ssel: no I think this is better :+1:

PR #20, opened 2021-01-20
Envproxy-only chart for workspace providers beta

deansheather:
# !!! DO NOT MERGE THIS PR !!!

This PR exists only for comments/review and diff. It should never be merged as it's basically a fork.

This is what coder-cli will use to deploy a resource pool. This will probably be eventually turned into a subchart to reduce duplication


values.yaml
```diff
@@ -20,6 +13,10 @@ podSecurityPolicyName: ""
 # resolution to work correctly.
 clusterDomainSuffix: ".svc.cluster.local"
 
+# imagePullPolicy -- Sets the policy for pulling a container image across all
+# services.
+imagePullPolicy: Always
```
f0ssel, line 18: In the spirit of prefixing everything, should we make a separate imagePullPolicy key for each service + environments?

deansheather: This will happen for Coder 2.0. I'm not sure if workspace providers will release before that, but if not this is just for dogfood.

values.yaml
```diff
@@ -182,40 +107,17 @@ envproxy:
...
+cemanager:
+  # cemangaer.accessURL -- The cemanager access URL that the envproxy will use
+  # to communicate with the cemanager. This must be the public cemanager URL
+  # because it's also used for redirecting unauthenticated users to the
+  # cemanager. This should be a full URL complete with protocol and no trailing
+  # slash.
+  accessURL: ""
+  # cemanager.token -- Set by coder-cli during the initial installation. Don't
+  # set this manually (even to empty string) unless you know what you're doing.
+  # token: ""
```
deansheather, line 120: this is envproxy.token

values.yaml
```diff
@@ -71,10 +71,9 @@ devurls:
 # connections to environments such as IDE or shell sessions.
 envproxy:
   # envproxy.accessURL -- The URL reported to cemanager. Must be accessible by
-  # cemanager and all users who can use this workspace provider. Derived from
-  # ingress.host if not set. This should be a full URL, complete with protocol
-  # and trailing "/proxy" (no trailing slash).
```
f0ssel: nitpick: maybe we use `wsp.coder.com` or `example.coder.com` to keep it generic, since a lot of the wsp docs don't reference the proxy directly.

f0ssel (approved):

f0ssel (approved):

PR #21, opened 2021-01-21
fix: pass namespace env into migrations init container and wait for pg

coadler:
Failed to see this due to a bug where the init container was exiting with 0 after failing.


tychoish (approved):
I wish that things like this were parameterized and passed as arguments rather than envvars, but this is fine

PR #22, opened 2021-01-25
fix: change readiness/liveness probes to reasonable intervals

coadler:
Now that we don't have to account for migrations while the probes are
running, we're free to set these back to reasonable times so it doesn't
take ages for cemanager/envproxy to be detected as ready.

### DIFF
--- a/templates/cemanager.yaml
+++ b/templates/cemanager.yaml
@@ -136,26 +136,18 @@ spec:
                 - curl
                 - -s
                 - localhost:8080/cem-healthz
-            initialDelaySeconds: 30
-            failureThreshold: 12
-            # Poll the manager every 20 seconds. If we fail 12 times (4
-            # minutes) the manager is most likely in a bad spot. It's possible
-            # for migrations to take extended periods of time, so we need to
-            # account for that.
-            periodSeconds: 20
+            initialDelaySeconds: 10
+            failureThreshold: 7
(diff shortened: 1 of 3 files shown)


templates/cemanager.yaml
```diff
@@ -136,26 +136,18 @@ spec:
...
                 - -s
                 - localhost:8080/cem-healthz
-            initialDelaySeconds: 30
-            failureThreshold: 12
-            # Poll the manager every 20 seconds. If we fail 12 times (4
-            # minutes) the manager is most likely in a bad spot. It's possible
-            # for migrations to take extended periods of time, so we need to
-            # account for that.
-            periodSeconds: 20
+            initialDelaySeconds: 10
```
sreya, line 139: do we even need this now? startup should be roughly instant right?

sreya (approved):

templates/cemanager.yaml
coadler: For some reason I still see pods fail to connect to pg for a few seconds as it first starts up, even if pg is available the entire time.
```
$ klo -n coder cemanager-98674fc85-mrc52 cemanager
Waiting on db
172.28.48.3:5432 - no response
172.28.48.3:5432 - no response
172.28.48.3:5432 - accepting connections
...
```
I'm guessing its some Istio thing.

PR #25, opened 2021-02-04
feat: support passing splunk logging configuration

tychoish:


values.yaml
```diff
@@ -289,3 +293,11 @@ logging:
...
   # empty string to disable.
   json: ""
+
+  ## Splunk logging configuration -- Coder can send logs directly to
+  # Splunk, in addition to file-based output, if these values are
+  # configured. The channel is optional, and this logging is disabled
+  # if either the URL and Token are set to the empty string.
+  splunk_url: ""
+  splunk_token: ""
+  splunk_channel: ""
```
coadler, line 303: I think having these under splunk key like
```
  splunk:
    url: ""
    ...`
```
object would make more sense

values.yaml
```diff
@@ -289,3 +293,11 @@ logging:
   # logging.json -- Where to send logs that are formatted as JSON. Set to an
   # empty string to disable.
   json: ""
+
+  ## Splunk logging configuration -- Coder can send logs directly to
```
coadler, line 297: These are intended to be 
```
<yaml key> -- comment string
```
where the yaml key matches up with the object path of what it's documenting

README.md
```diff
@@ -14,55 +14,63 @@ Coder!
...
-| timescale.resources.requests.storage   | string | Specifies the size of the volume claim for persisting the database.                                                                                                                                                                                                                                                                          | `"10Gi"`                                                                                                                      |
+| Key                                    | Type   | Description                                                                                                                                                                                                                                                                                                                                        | Default                                                                                                                       |
```
jawnsy, line 17: is the README generated from somewhere? otherwise this seems like a lot of changes, and i haven't read through this

jawnsy (approved):
LGTM. As a general thought, it seems like we should these to secrets or a ConfigMap, but definitely not now/not in this change

coadler (approved):

README.md
tychoish: yup! there's a script that does it.

PR #27, opened 2021-02-21
feat: allow specifying annotations for deployments

coadler:


values.yaml
```diff
@@ -304,3 +304,5 @@ logging:
     # logging.splunk_channel -- Optional. Specify the channel that you'd
     # like to associate with all messages.
     channel: ""
+
+deploymentAnnotations: {}
```
cmoog, line 308: probably deserves a comment because the language here can get a big confusing. For example: is `deployment` referring to the Kubernetes resource or the fact that it's a Coder service?

cmoog (approved):

values.yaml
coadler: Didn't want to document this just yet since I'm not convinced I want to commit to this name/structure just yet. Currently this means just Coder created deployments (i.e. Coder services, but calling it `serviceAnnotations` could mean it applies to k8s service specs) and the pods created by them. I potentially see adding more configurability to this like annotating deployments and pods differently, annotating each service independently, etc. As it is currently I'm using this to configure Istio sidecars, so I'll flesh it out a bit more once I get some feedback.

PR #28, opened 2021-02-21
feat: separate ingress object from default nginx deployment

coadler:
* deprecates the old ugly way of doing ingress annotations
* switches all the health checks to use k8s native http requests so cloud providers can detect them
* allows the ingress kind to be created separately from the built-in ingress-nginx
* fixes deployments using `ingress-gce` and azure AGIC

Further, some work will be need to be done to see what configurations work with each cloud provider. So far I've tested and have working deploys using ingress-gce and azure AGIC, but I'm sure there's plenty more to test.


README.md
```diff
@@ -48,11 +51,12 @@ Coder!
...
+| logging.splunk.token                   | string | The Splunk HEC collector token.                                                                                                                                                                                                                                                                                                                    | `""`                                                                                                                          |
+| logging.splunk.url                     | string | The Splunk HEC collector endpoint.                                                                                                                                                                                                                                                                                                                 | `""`                                                                                                                          |
```
sreya, line 59: :+1:

templates/ingress.yaml
```diff
@@ -1,6 +1,6 @@
 # This configuration file is ported from the nginx repo:
 # See https://raw.githubusercontent.com/kubernetes/ingress-nginx/master/deploy/static/mandatory.yaml
-{{- if eq .Values.ingress.useDefault  true}}
+{{- if .Values.ingress.useDefault }}
```
sreya, line 3: :+1:

templates/ingress.yaml
```diff
@@ -381,5 +393,32 @@ spec:
...
+  - host: {{ .Values.devurls.host | quote }}
+    http:
+      paths:
+      - path: /*
+        backend:
+          serviceName: envproxy
+          servicePort: 8080
 {{- end }}
+    {{- end }}
+{{- end }}
```
sreya, line 423: the indentations for the `{{ if` are tripping me out. Is it possible to format it better?

sreya (approved):
Looks solid

templates/ingress.yaml
coadler: kinda awkward cuz i tried to keep them at the indentation of what they wrapped. fixed tho

PR #40, opened 2021-03-10
chore: embed template directly

jawnsy:
Make documentation using an embedded template, so that we can
customize output without using a forked version of helm-docs.


Makefile
```diff
@@ -19,3 +20,10 @@ lint/helm:
 	# TODO(jawnsy): enable --strict once we fix the warnings
 	helm lint .
 .PHONY: lint/helm
+
+docs: README.md
+.PHONY: docs
+
+README.md: README.md.gotmpl
+	@echo "--- Generating documentation"
+	helm-docs --template-files=$<
```
coadler, line 29: Should we still use prettier to format the table after generation?

coadler (approved):

Makefile
jawnsy: yeah, will do in a follow-up though, because then I also have to install prettier :sweat:

PR #41, opened 2021-03-10
ci: additional linting for Kubernetes YAML files

jawnsy:
* Render example Values files to Kubernetes YAML outputs,
  then lint them using kube-linter


examples/release.yaml
```diff
@@ -0,0 +1,13 @@
...
+dashboard:
+  image: docker.io/coderenvs/dashboard:1.16.1
+envproxy:
+  image: docker.io/coderenvs/coder-service:1.16.1
+timescale:
+  image: docker.io/coderenvs/timescale:1.16.1
+envbox:
+  image: docker.io/coderenvs/envbox:1.16.1
+envbuilder:
+  image: docker.io/coderenvs/envbuilder:1.16.1
```
deansheather, line 13: These should just be fake image URLs, I doubt it matters if they are real or not. Saves us the effort of updating this every release.

I'd probably rename this to `images.yaml`, or merge into `kind.values.yaml`.

examples/kind/kind.values.yaml
```diff
@@ -0,0 +1,27 @@
+# Values for a small local test deployment using Kind
```
jawnsy, line 1: I need to add some more examples, but will do that in a follow-up to keep this PR small

.gitignore
```diff
@@ -0,0 +1 @@
+/build
```
deansheather, line 1: While we're on this topic, we should add `.github`, `Makefile`, `examples`, `script` to helmignore

scripts/test_helm.sh
```diff
@@ -0,0 +1,30 @@
...
+
+check_dependencies \
+  git \
+  helm
+
+PROJECT_ROOT="$(git rev-parse --show-toplevel)"
+
+EXAMPLES=" \
+  kind \
+"
```
deansheather, line 18: better suited for a bash array

deansheather (approved):

examples/release.yaml
jawnsy: fair point, will rename. I didn't want to merge it because my thinking was that we'd have other values files later, that share a common list 'o images

.gitignore
jawnsy: oooh, great point, whoopsie

scripts/test_helm.sh
jawnsy: excellent point, will change

.github/workflows/build.yml
```diff
@@ -32,3 +34,13 @@ jobs:
...
         if: always()
         run: make lint/shellcheck
+
+      - name: Lint Kubernetes Templates
+        if: always()
+        run: make lint/kubernetes
+        # TODO(jawnsy): fix linter warnings and remove this
+        continue-on-error: true
+
+      - name: Package Helm Chart
```
jawnsy, line 44: @deansheather Updated the .helmignore and added a build step to package up the chart and show the list of files

examples/images.yaml
```diff
@@ -0,0 +1,14 @@
+# Dummy image tags
+# Image tags from the latest release
+cemanager:
+  image: gcr.io/coder/coder-service:2.0.0
```
jawnsy, line 3: @deansheather Renamed release -> images and made dummy version tags. This doesn't matter for any of the linters currently in-use, but `kube-score` does check that we're not using the `default/latest` tag.

PR #42, opened 2021-03-11
Deprecate env whitelist

deansheather:


f0ssel (approved):
@nathanpotter If you can do a once over on this messaging I'd appreciate it. Won't be the only place, there will be places in docs explaining as well, but still wanna get an OK

nathanpotter (approved):

PR #43, opened 2021-03-12
Add env service account env var

deansheather:


templates/cemanager.yaml
```diff
@@ -120,11 +120,6 @@ spec:
                 secretKeyRef:
                   name: built-in-envproxy-token
                   key: token
-              # ENVIRONMENT_SERVICE_ACCOUNT is the service account to assign
```
deansheather, line 123: this comment is in another place already (environment.yaml) so it's kinda redundant anyways

f0ssel (approved):

f0ssel:
So next is to read from this env, and send that in the resource pool update payload, correct?

PR #45, opened 2021-03-18
chore: add coder.deployment label for PostgreSQL

jawnsy:
Add a coder.deployment label to be consistent with other components,
to simplify our ce-diagnostics script.

# How I tested this

I used `make deploy` and then verified that the label appears on the timescale-0 pod:

```shell-session
[coder@jawnsy-m enterprise-helm]$ kubectl describe pod timescale-0
Name:         timescale-0
Namespace:    coder-jawnsy-m
Priority:     0
Node:         gke-dev-2-default-pool-cba3c583-0qyu/10.1.0.62
Start Time:   Thu, 18 Mar 2021 01:00:35 +0000
Labels:       app=timescale
              coder.deployment=postgresql
              controller-revision-hash=timescale-994689566
              statefulset.kubernetes.io/pod-name=timescale-0
Annotations:  cni.projectcalico.org/podIP: 10.8.3.226/32
```

### DIFF
--- a/templates/timescale.yaml
+++ b/templates/timescale.yaml
@@ -7,7 +7,7 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
     app: timescale
-    coder.deployment: postgresql
+    coder.deployment: timescale
 spec:
   ports:
     - port: 5432
@@ -34,7 +34,7 @@ spec:
     metadata:
       labels:
         app: timescale
-        coder.deployment: postgresql
+        coder.deployment: timescale
       annotations:
       {{- range $key, $value := .Values.deploymentAnnotations }}
         {{ $key }}: {{ $value | quote }}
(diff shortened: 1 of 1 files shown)


templates/timescale.yaml
```diff
@@ -7,6 +7,7 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
     app: timescale
+    coder.deployment: postgresql
```
jawnsy, line 10: I called this postgresql because, well, the pod is actually postgresql... but we can also call it timescale if y'all think that's better

deansheather: We should call it timescale to match the app label and the deployment name

deansheather, line 10: We should call it timescale to match the app label and the deployment name

deansheather (approved):

templates/timescale.yaml
jawnsy: So we're gonna stick with this lie forever? :rofl: Fair enough I guess

PR #54, opened 2021-04-12
Remove hardcoded /proxy path from built-in wsp

f0ssel:
I originally didn't know where this was coming from, and thought that we just had the data in the DB from 1.16 previously, but when digging around I saw that we still hard code the `/proxy` path for built-in WSPs. This will set it to just the hostname and will apply on deployment.

### DIFF
--- a/templates/_common.tpl
+++ b/templates/_common.tpl
@@ -100,7 +100,7 @@ tolerations:
     {{- else -}}
     http://
     {{- end -}}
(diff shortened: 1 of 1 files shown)


vapurrmaid:
@deansheather @f0ssel does this change mean we need to deprecate `1.17.2` and release `1.17.3` ?

deansheather (approved):

PR #56, opened 2021-04-19
Initial json schema, still very loose

Emyrk:
This is a start.

@coadler @deansheather If you guys want to add tighter validation, I can hop on a call and peer code this json schema to enforce minimums, regex patterns, exclusive fields, etc

# Future Work

- Tighter validation
- More descriptions

### DIFF
--- a/values.schema.json
+++ b/values.schema.json
@@ -388,17 +388,29 @@
     },
     "definitions": {
         "kube_resource": {
-            "$comment": "Can this also be a number? Add regex to the strings to be valid resource strings",
-            "type": "string"
+            "type": "object",
+            "additionalProperties": false,
+            "properties": {
+                "storage": {
+                    "$comment": "Can this also be a number? Add regex to the strings to be valid resource strings",
+                    "type": "string"
+                },
+                "cpu": {
+                    "type": "string"
+                },
+                "memory": {
+                    "type": "string"
+                }
+            }
         },
         "kube_resources": {
             "type": "object",
             "additionalProperties": false,
             "properties": {
-                "cpu": {
+                "requests": {
                     "$ref": "#/definitions/kube_resource"
                 },
-                "memory": {
+                "limits": {
                     "$ref": "#/definitions/kube_resource"
                 }
             }
--- a/values.schema.json
+++ b/values.schema.json
@@ -38,12 +38,12 @@
         },
         "ingress": {
             "type": "object",
+            "required": [
+                "enable",
+                "useDefault"
+            ],
             "properties": {
                 "additionalProperties": false,
-                "required": [
-                    "enable",
-                    "useDefault"
-                ],
                 "enabled": {
                     "$comment": "TODO: Check default",
                     "type": "boolean",
@@ -168,7 +168,7 @@
                     "type": "string"
                 },
                 "resources": {
-                    "$ref": "#definitions/kube_resources"
+                    "$ref": "#/definitions/kube_resources"
                 }
             }
         },
@@ -183,7 +183,7 @@
                     "type": "string"
                 },
                 "resources": {
-                    "$ref": "#definitions/kube_resources"
+                    "$ref": "#/definitions/kube_resources"
                 }
             }
         },
@@ -204,7 +204,7 @@
                     "type": "string"
                 },
                 "resources": {
-                    "$ref": "#definitions/kube_resources"
+                    "$ref": "#/definitions/kube_resources"
                 },
                 "terminationGracePeriodSeconds": {
                     "type": "number",
@@ -220,7 +220,7 @@
                     "type": "string"
                 },
                 "resources": {
-                    "$ref": "#definitions/kube_resources"
+                    "$ref": "#/definitions/kube_resources"
                 }
             }
         },
@@ -347,7 +347,7 @@
             }
         },
         "serviceTolerations": {
-            "$ref": "#definitions/kube_tolerations"
+            "$ref": "#/definitions/kube_tolerations"
         },
         "logging": {
             "type": "object",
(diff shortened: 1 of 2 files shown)


values.schema.json
```diff
@@ -0,0 +1,479 @@
...
+    "title": "Coder Enterprise Helm Files",
+    "description": "TODO",
+    "type": "object",
+    "additionalProperties": false,
+    "$comment": "Set all the required fields",
+    "required": [],
+    "properties": {
+        "storageClassName": {
+            "type": "string",
+            "$comment": "TODO: Is this always a string? Is null an ok field? Can it be omitted?",
```
deansheather, line 13: Always string, can be left empty to use the default on the cluster.

coadler (commented):
Could you add a make target for generating this?

values.schema.json
```diff
@@ -0,0 +1,479 @@
...
+        },
+        "serviceType": {
+            "type": "string",
+            "default": "ClusterIP",
+            "$comment": "TODO: Check the enum list is complete",
+            "enum": [
+                "ClusterIP",
+                "NodePort",
+                "LoadBalancer",
+                "ExternalName"
```
deansheather, line 24: ExternalName isn't supported by our chart because you need to set extra fields for it.

values.schema.json
```diff
@@ -0,0 +1,479 @@
...
+            ]
+        },
+        "podSecurityPolicyName": {
+            "type": "string",
+            "default": ""
+        },
+        "clusterDomainSuffix": {
+            "$comment": "TODO: Should we apply any regex here?",
```
deansheather, line 32: should start with period and be a valid hostname

jawnsy:
@coadler My thought is that we should invert things, we would update the schema (with descriptions) and use the schema to generate our docs (instead of helm-docs) and optionally the values file (with comments, if we want to keep those). I think just the values files without descriptions would be OK too, personally, because someone can run `helm lint` to validate their values file (which they cannot do today). Structured JSON is a lot easier for programs to work with as compared to parsing YAML comments :vomiting_face:

values.schema.json
```diff
@@ -0,0 +1,479 @@
...
+            "type": "string"
+        },
+        "ingress": {
+            "type": "object",
+            "required": [],
+            "properties": {
+                "additionalProperties": false,
+                "enabled": {
+                    "$comment": "TODO: Check default",
```
deansheather, line 41:
```suggestion
```

values.schema.json
```diff
@@ -0,0 +1,479 @@
...
+                    "type": "boolean",
+                    "default": true
+                },
+                "useDefault": {
+                    "$comment": "TODO: Check default",
```
deansheather, line 46:
```suggestion
```

values.schema.json
```diff
@@ -0,0 +1,479 @@
+{
```
deansheather, line 1: Either all fields should have default+descriptions set, or none should IMO. I'm leaning towards not having defaults+descriptions set and relying on the values file in the repo on being this source of info

values.schema.json
```diff
@@ -0,0 +1,479 @@
...
+                        "verify-ca",
+                        "verify-full"
+                    ],
+                    "default": "require"
+                }
+            }
+        },
+        "imagePullPolicy": {
+            "type": "string",
+            "default": "Always"
```
deansheather, line 153: Should fill in the enum for this, I think it's just `Always` `IfNotPresent` and `Never`.

values.schema.json
```diff
@@ -0,0 +1,479 @@
...
+                    "type": "string"
+                },
+                "replicas": {
+                    "type": "number"
+                },
+                "image": {
+                    "type": "string"
+                },
+                "resources": {
+                    "$ref": "#/definitions/kube_resources"
```
deansheather, line 169: nice 👍

values.schema.json
```diff
@@ -0,0 +1,479 @@
...
+                                "type": "integer",
+                                "minimum": 0
+                            }
+                        }
+                    }
+                },
+                "nodeSelector": {
+                    "type": "object",
+                    "patternProperties": {
+                        "^.": {
```
deansheather, line 307:
```suggestion
                        ".*": {
```

values.schema.json
```diff
@@ -0,0 +1,479 @@
...
+                            "$comment": "All values should be strings",
```
deansheather, line 308:
```suggestion
```

values.schema.json
```diff
@@ -0,0 +1,479 @@
...
+                            "null"
+                        ],
+                        "enum": [
+                            "NoSchedule",
+                            "PreferNoSchedule",
+                            "NoExecute",
+                            null
+                        ],
+                        "default": null
+                    },
```
deansheather, line 470: Are these allowed to be null?

values.schema.json
```diff
@@ -0,0 +1,479 @@
...
+                },
+                "limits": {
+                    "$ref": "#/definitions/kube_resource"
+                }
+            }
+        },
+        "kube_tolerations": {
+            "type": "array",
+            "items": {
+                "additionalProperties": false,
```
deansheather, line 424: `"type": "object"`?

values.schema.json
```diff
@@ -0,0 +1,479 @@
...
+                "image": {
+                    "type": "string"
+                }
+            }
+        },
+        "environments": {
+            "type": "object",
+            "additionalProperties": false,
+            "properties": {
+                "tolerations": {
```
deansheather, line 247: This should just use the definition you made

values.schema.json
```diff
@@ -0,0 +1,479 @@
...
+                        "channel": {
+                            "type": "string"
+                        }
+                    }
+                }
+            }
+        },
+        "deploymentAnnotations": {
+            "type": "object"
+        }
```
deansheather, line 390: This is just `[string]: string` like `nodeSelector`

values.schema.json
```diff
@@ -0,0 +1,479 @@
...
+    },
+    "definitions": {
+        "kube_resource": {
+            "type": "object",
+            "additionalProperties": false,
+            "properties": {
+                "storage": {
+                    "$comment": "Can this also be a number? Add regex to the strings to be valid resource strings",
```
deansheather, line 398: Not sure if numbers are supported, but we should just keep it string only anyways to avoid ambiguity

values.schema.json
deansheather: or generating values.yaml from this

Emyrk:
This is way too old

PR #59, opened 2021-04-22
feat: Expose ports for WebRTC TURN server 

kylecarbs:
This is behind a deployment feature flag, so will not provide any security concerns.

### DIFF
--- a/templates/cemanager.yaml
+++ b/templates/cemanager.yaml
@@ -155,8 +155,16 @@ spec:
             failureThreshold: 7
             periodSeconds: 10
 {{- include "coder.resources" .Values.cemanager.resources | indent 10 }}
-{{- include "coder.volumeMounts" . | indent 10 }}
-{{- include "coder.volumes" . | indent 6 }}
+{{- if .Values.ingress.tls.enable }}
+          volumeMounts:
+            - mountPath: /etc/coder/certificates
+              name: tls
+              readOnly: true
+      volumes:
+        - name: tls
+          secret:
+            secretName: {{ .Values.ingress.tls.hostSecretName | quote }}
+{{-end}}
 ---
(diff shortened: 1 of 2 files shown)


templates/cemanager.yaml
```diff
@@ -113,6 +113,10 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
+            - name: POD_IP
+              valueFrom:
+                fieldRef:
+                  fieldPath: status.podIP
```
f0ssel, line 119: So this is so a replica behind an LB can connect to other replicas in the same cluster. What happens if this IP is a pod in a different cluster (like in a future where coderd is running multi-region)?

templates/ingress.yaml
```diff
@@ -37,6 +37,7 @@ metadata:
     app.kubernetes.io/part-of: ingress-nginx
 data:
   22: "{{ .Release.Namespace }}/envproxy:2222"
+  5349: "{{ .Release.Namespace }}/cemanager:5349"
```
f0ssel, line 40: Did we find out we _don't_ need any proxy protocol here?

f0ssel (approved):
Lgtm from what I can tell

templates/ingress.yaml
kylecarbs: We don't!

It's kinda a separate issue requiring the PROXY proto TBH... it allows us to fully forward external IPs which we should have for auditing, but isn't a neeeeed right now.

templates/cemanager.yaml
kylecarbs: Different clusters require a unique hostname setup for each. We can propagate this error really nicely to the user too!

PR #62, opened 2021-04-28
chore: edit Helm chart descriptions

khorne3:
- edits to descriptions used in `values.yaml`
- edit, minor reorg to README.md

### DIFF
--- a/README.md.gotmpl
+++ b/README.md.gotmpl
@@ -5,7 +5,7 @@ since the default branch (master) may contain updates for an unreleased version
 of Coder.
 
 During Coder's release process, we inject version numbers and image references
-(i.e., `docker.io/coderenvs/coder-service:1.17.2`), then publish the result to
+(e.g., `docker.io/coderenvs/coder-service:1.17.2`), then publish the result to
 our [Helm repository](https://helm.coder.com).
 
 When installing Coder, use the official packages that you can obtain as follows:
--- a/README.md.gotmpl
+++ b/README.md.gotmpl
@@ -1,25 +1,39 @@
 # Coder Helm Chart
 
-⚠️ **WARNING**: The default branch (master) may contain updates for an
-unreleased version of Coder. The current state of the repository may not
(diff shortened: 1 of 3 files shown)


README.md.gotmpl
```diff
@@ -1,25 +1,39 @@
 # Coder Helm Chart
 
-⚠️ **WARNING**: The default branch (master) may contain updates for an
-unreleased version of Coder. The current state of the repository may not
-represent the current release.
+⚠️ **WARNING**: This repository may not represent the current Coder release,
```
jawnsy, line 3: nice! I really like this phrasing

README.md.gotmpl
```diff
@@ -1,25 +1,39 @@
...
+since the default branch (master) may contain updates for an unreleased version
+of Coder.
 
 During Coder's release process, we inject version numbers and image references
-(e.g. `docker.io/coderenvs/coder-service:1.17.2`), then publish the result
-to our [Helm repository](https://helm.coder.com).
+(i.e., `docker.io/coderenvs/coder-service:1.17.2`), then publish the result to
```
jawnsy, line 8: I used e.g. rather than i.e. because there's multiple different image references: docker.io/coderenvs/coder-service:1.17.2, docker.io/coderenvs/dashboard:1.17.2, docker.io/coderenvs/envproxy:1.17.2 etc - but am happy for the correction if I was using it incorrectly! 

My understanding was that i.e. should be used in place of a phrase like "in other words", in which case saying "image references (in other words ...)" does not seem correct since it just mentions one type of image reference and not all of them

jawnsy (approved):
Nice changes, thank you Katie!

README.md.gotmpl
khorne3: @jawnsy I [double-checked](https://www.grammarly.com/blog/know-your-latin-i-e-vs-e-g/) and you're correct. Thanks for keeping me honest!

PR #63, opened 2021-04-28
update ingress external traffic policy

ericpaulsen:
changing ingress `externalTrafficPolicy` value from `Local` to `Cluster`, as it fixes the issue with AWS ELB not forwarding traffic to a node that has the nginx ingress instance running, as documented [here](https://app.clubhouse.io/coder/story/9632/eks-load-balancer-is-running-but-status-says-it-s-out-of-service-and-doesn-t-route-traffic).

a customer found this solution from this [GitLab issue](https://gitlab.com/gitlab-org/charts/gitlab/-/issues/939). only issue is that it's making the IP addresses of the routed traffic private, which removes the ability for our audit logs to properly identify people.

submitting this PR to open up the discussion.

### DIFF
--- a/templates/ingress.yaml
+++ b/templates/ingress.yaml
@@ -315,7 +315,7 @@ metadata:
     {{ $key }}: {{ $value | quote }}
   {{- end }}
 spec:
-  externalTrafficPolicy: Local
+  externalTrafficPolicy: Cluster
   type: LoadBalancer
   loadBalancerIP: {{ .Values.ingress.loadBalancerIP | quote }}
   selector:


f0ssel:
We can try adding https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#use-forwarded-headers and see if that makes the right IP come through. I can try it out and get back to you.

ericpaulsen:
Where would that be added, in the `ingress` spec?

coadler:
> Where would that be added, in the `ingress` spec?

It's in the configmap, but I think that's reasonable for us to add in a separate pr, since it needs some decent configurability

coadler (approved):

ericpaulsen:
Would it be wise to hold off on merging this PR until the forwarded header PR is ready? Might be best to avoid customers installing via helm & them only seeing private IPs before we've fixed it.

ericpaulsen:
> Would it be wise to hold off on merging this PR until the forwarded header PR is ready? Might be best to avoid customers installing via helm & them only seeing private IPs before we've fixed it.

On second thought, I'm going to merge this in, as I think it's best to fix the near-term ELB issue for customers.

bpmct:
Sanity check - this will only effect charts in 1.19+, correct?

coadler:
> Sanity check - this will only effect charts in 1.19+, correct?

Correct

PR #65, opened 2021-05-04
chore: Remove envbuilder

kylecarbs:
We should be cautious merging this. We'll need to document the removal of envbuilder, as I'd assume helm will throw an error if a customer keeps this value.

### DIFF
--- a/templates/cemanager.yaml
+++ b/templates/cemanager.yaml
@@ -100,11 +100,6 @@ spec:
               value: {{ .Values.logging.json | quote }}
             - name: STACKDRIVER_LOG
               value: {{ .Values.logging.stackdriver | quote }}
-              # ENVBUILDER_IMAGE describes the image used to build
-              # user images and populate them with coder assets
-              # (such as code-server).
-            - name: ENVBUILDER_IMAGE
-              value: {{ .Values.envbuilder.image | quote }}
               # ENVBOX_IMAGE describes the image used to provide
               # additional features to users for running applications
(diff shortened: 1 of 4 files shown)


f0ssel (approved):
No errors occur if you specify extra data in values for helm charts, so this shouldn't break anything.

templates/cemanager.yaml
```diff
@@ -100,11 +100,6 @@ spec:
               value: {{ .Values.logging.json | quote }}
             - name: STACKDRIVER_LOG
               value: {{ .Values.logging.stackdriver | quote }}
-              # ENVBUILDER_IMAGE describes the image used to build
-              # user images and populate them with coder assets
-              # (such as code-server).
-            - name: ENVBUILDER_IMAGE
```
jawnsy, line 106: I guess cemanager did some magic to use this image (e.g. creating a separate pod spec using envbuilder) rather than us installing it via Helm?

jawnsy (approved):
Given @f0ssel's comment, it seems like this is an easily-reversible decision, and it makes sense that we wouldn't need to provide this image anymore since we're not using it

templates/cemanager.yaml
kylecarbs: Yup!

jawnsy:
@kylecarbs Are you planning to merge this, or are we waiting until the next release? Since the release cut is coming soon

kylecarbs:
I'll merge!

PR #66, opened 2021-05-04
feat: configurable ingress externalTrafficPolicy and deploy kind

coadler:
Allows configuring two new variables

`ingress.service.externalTrafficPolicy`: It was found that setting this to `Local` causes issues in EKS, and it was changed to `Cluster` in https://github.com/cdr/enterprise-helm/pull/63. This changes the default back to `Local` since this is the most sensible default, and makes it configurable.

`ingress.deployment.kind`: In deployments that use `externalTrafficPolicy: Cluster`, the user's IP becomes masked by kube-proxy. This allows ingress-nginx to be deployed as a `DaemonSet` to prevent this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,12 +67,15 @@ form](https://coder.com/contact) or send an email to support@coder.com.
 | imagePullPolicy | string | Sets the policy for pulling a container image across all services. | `"Always"` |
 | ingress.additionalAnnotations | list | Deprecated. Please use `ingress.annotations`. | `[]` |
 | ingress.annotations | object | Additional annotations to be used when creating the ingress. These only apply to the Ingress Kubernetes kind. The annotations can be used to specify certificate issuers or other cloud provider-specific integrations. | `{}` |
+| ingress.deployment | object | Options used to configure the Deployment of the default ingress controller. | `{"kind":"Deployment"}` |
+| ingress.deployment.kind | string | How to deploy the built-in ingress. Can be "Deployment" or "DaemonSet". | `"Deployment"` |
 | ingress.enable | bool | If set to true, a Coder compatible ingress kind will be created. You can configure it with `ingress.annotations` below. | `true` |
 | ingress.host | string | The hostname to use for accessing the platform. This can be left blank, and the user can still access the platform from the external IP or a DNS name that resolves to the external IP address. | `""` |
 | ingress.loadBalancerIP | string |  | `""` |
 | ingress.podSecurityPolicyName | string | The name of the pod security policy the built-in ingress controller should abide. It should be noted that the ingress controller requires the `NET_BIND_SERVICE` capability, privilege escalation, and access to privileged ports to successfully deploy. Ignored if `ingress.useDefault` is false. | `""` |
-| ingress.service | object | Options related to the ingress Kubernetes Service object. | `{"annotations":{}}` |
+| ingress.service | object | Options used to configure the LoadBalancer Service used in the default ingress controller. | `{"annotations":{},"externalTrafficPolicy":"Local"}` |
 | ingress.service.annotations | object | Additional annotations to add to the Service object. For example, to make the ingress spawn an internal load balancer: annotations:  cloud.google.com/load-balancer-type: "Internal" | `{}` |
+| ingress.service.externalTrafficPolicy | string | Denotes whether to route external traffic to node-local or cluster-wide endpoints. "Local" routes external traffic directly to the correct node, while "Cluster" evenly distributes traffic across all nodes, but masks source IPs. | `"Local"` |
 | ingress.tls | object | TLS options for the ingress. The hosts used for the tls configuration come from the ingress.host and the devurls.host variables. If those don't exist, then the TLS configuration will be ignored. | `{"devurlsHostSecretName":"","enable":false,"hostSecretName":""}` |
 | ingress.tls.devurlsHostSecretName | string | The secret to use for the devurls.host hostname. | `""` |
 | ingress.tls.enable | bool | Enables the tls configuration. | `false` |
--- a/templates/ingress.yaml
+++ b/templates/ingress.yaml
@@ -203,7 +203,7 @@ subjects:
     namespace: {{ .Release.Namespace | quote }}
 ---
 apiVersion: apps/v1
-kind: Deployment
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -67,12 +67,15 @@ form](https://coder.com/contact) or send an email to support@coder.com.
 | imagePullPolicy | string | Sets the policy for pulling a container image across all services. | `"Always"` |
 | ingress.additionalAnnotations | list | Deprecated. Please use `ingress.annotations`. | `[]` |
 | ingress.annotations | object | Additional annotations to be used when creating the ingress. These only apply to the Ingress Kubernetes kind. The annotations can be used to specify certificate issuers or other cloud provider-specific integrations. | `{}` |
+| ingress.deployment | object | Options used to configure the Deployment of the default ingress controller. | `{"kind":"Deployment"}` |
+| ingress.deployment.kind | string | How to deploy the built-in ingress. Can be "Deployment" or "DaemonSet". | `"Deployment"` |
```
jawnsy, line 71: Not sure I understand this - in the case where we want a cluster-wide ingress controller, doesn't it make more sense for users to deploy their own and set `ingress.enable` and `ingress.useDefault` to false?

README.md
```diff
@@ -67,12 +67,15 @@ form](https://coder.com/contact) or send an email to support@coder.com.
...
 | ingress.enable | bool | If set to true, a Coder compatible ingress kind will be created. You can configure it with `ingress.annotations` below. | `true` |
 | ingress.host | string | The hostname to use for accessing the platform. This can be left blank, and the user can still access the platform from the external IP or a DNS name that resolves to the external IP address. | `""` |
 | ingress.loadBalancerIP | string |  | `""` |
 | ingress.podSecurityPolicyName | string | The name of the pod security policy the built-in ingress controller should abide. It should be noted that the ingress controller requires the `NET_BIND_SERVICE` capability, privilege escalation, and access to privileged ports to successfully deploy. Ignored if `ingress.useDefault` is false. | `""` |
-| ingress.service | object | Options related to the ingress Kubernetes Service object. | `{"annotations":{}}` |
+| ingress.service | object | Options used to configure the LoadBalancer Service used in the default ingress controller. | `{"annotations":{},"externalTrafficPolicy":"Local"}` |
```
jawnsy, line 76: When does one set `externalTrafficPolicy` here versus `ingress.service.externalTrafficPolicy`?

README.md
```diff
@@ -67,12 +67,15 @@ form](https://coder.com/contact) or send an email to support@coder.com.
...
 | ingress.service.annotations | object | Additional annotations to add to the Service object. For example, to make the ingress spawn an internal load balancer: annotations:  cloud.google.com/load-balancer-type: "Internal" | `{}` |
+| ingress.service.externalTrafficPolicy | string | Denotes whether to route external traffic to node-local or cluster-wide endpoints. "Local" routes external traffic directly to the correct node, while "Cluster" evenly distributes traffic across all nodes, but masks source IPs. | `"Local"` |
```
jawnsy, line 78: Reading the [Kubernetes documentation for external load balancers](https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/#preserving-the-client-source-ip), it says this:

> denotes if this Service desires to route external traffic to node-local or cluster-wide endpoints. There are two available options: Cluster (default) and Local. Cluster obscures the client source IP and may cause a second hop to another node, but should have good overall load-spreading. Local preserves the client source IP and avoids a second hop for LoadBalancer and NodePort type services, but risks potentially imbalanced traffic spreading.

I'm unclear though, if the ingress is proxying for us, doesn't it add an X-Forwarded-For header that we can read for the client's IP address? In which case Cluster seems to make more sense, since it means we have good overall load sharing (per the above description)

jawnsy (approved):

coadler:
[ch11989]

ericpaulsen (approved):

README.md
coadler: You don't. Tbh this is just an unfortunate part of helm-docs, since it tries to show the default of the entire object.

README.md
coadler: I think this would be optimal, but doing so currently is not that easy. Right now we expect customers to use our ingress unless it's not possible to do so. At a certain point we start reimplementing the `ingress-nginx` helm chart, but I don't see a good alternative atm

README.md
coadler: What happens is if the traffic is proxied from a different node before hitting our ingress, we lose the original client IP. There isn't a way to preserve this through kube-proxy unfortunately.

README.md
jawnsy: Once we have the agent architecture & coderd work that @f0ssel and @kylecarbs are driving, it seems like we won't need path-based routing anymore, so then we can conceivably drop the ingress too, right? That would allow us to remove this stuff from our chart

coadler: Yea p2p could let us drop nginx-ingress for more cloud native solutions, but it would still be a breaking change. We would still need to deploy an Ingress, but the controller backing it can be much more flexable.

README.md
jawnsy: I still don't understand why we lose the original client IP. `kube-proxy` is doing iptables stuff, so I get why, if we try to look at the client IP from `net/http`'s RemoteAddr, it won't be correct. But what I'm saying is that, we can have a middleware somewhere extract the original client IP from the X-Forwarded-For header, and then update RemoteAddr? That's what the Chi RealIP middleware is doing: https://github.com/go-chi/chi/blob/f16f8704ccc32f63fbc20eb364c41050cf339860/middleware/realip.go#L14-L38

coadler: Here's a well written explanation of the issue: https://www.asykim.com/blog/deep-dive-into-kubernetes-external-traffic-policies

`kube-proxy` is only an issue if it forwards traffic *between* nodes

jawnsy: Oh, I guess this is because of the way traffic is routed to the _ingress controller_, which does need to see the client's IP address. I got confused and thought this was referring to the traffic routed from the ingress controller to the _cemanager_ service, which does not actually need to see the client IP address, since it can read the `X-Forwarded-For` header added by the ingress controller. Okay, that all makes sense then.

coadler: Yep! Apologies for the confusion

deansheather (approved):

PR #71, opened 2021-05-07
chore: update ingress-nginx image

coadler:
[ch12199]

### DIFF
--- a/templates/ingress.yaml
+++ b/templates/ingress.yaml
@@ -241,7 +241,7 @@ spec:
 {{- include "coder.serviceTolerations" . | indent 6 }}
       containers:
         - name: nginx-ingress-controller
-          image: quay.io/kubernetes-ingress-controller/nginx-ingress-controller:0.26.1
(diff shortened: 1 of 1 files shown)


deansheather (approved):

templates/ingress.yaml
```diff
@@ -259,8 +259,8 @@ spec:
                 - ALL
               add:
                 - NET_BIND_SERVICE
-            # www-data -> 33
-            runAsUser: 33
+            # www-data -> 101
+            runAsUser: 101
```
jawnsy, line 263: we don't store any files right? this won't affect any sort of filesystem permission things?

jawnsy (approved):

templates/ingress.yaml
coadler: they're all via configmaps it seems, didn't notice anything break in my testing

deansheather: no volumes, and i verified that the user is correct by pulling the image manually

jawnsy: cool, sounds good, thanks for checking :+1:

PR #77, opened 2021-05-12
feat: Remove dashboard image

kylecarbs:
**Should not be merged before 1.19 is cut.**

Removes the dashboard image. Requires a change to the ingress... directing all HTTP traffic to cemanager.


f0ssel (approved):

templates/ingress.yaml
```diff
@@ -372,9 +372,6 @@ metadata:
   {{- end }}
 spec:
 {{- include "coder.ingress.tls" . }}
-  backend:
-    serviceName: dashboard
```
deansheather, line 376: Shouldn't this become cemanager 8080

deansheather (approved):
1.19 has already been cut

templates/ingress.yaml
kylecarbs: This appears to be an unused property... but @coadler is on the blame so he'll let us know

coadler: This is the backend it sends traffic to when it matches none of the rules. It should probably be changed to cemanager

kylecarbs: Gotcha. Will update.

PR #81, opened 2021-05-13
fix: Reduce permissions required for deployment

kylecarbs:
From what I can tell, these are the only permissions we need.

I have very little knowledge of the implications this has... someone with greater context should ensure we'll have no problems here.

### DIFF
--- a/templates/rbac.yaml
+++ b/templates/rbac.yaml
@@ -97,4 +97,3 @@ roleRef:
   kind: Role
   name: coder
   apiGroup: rbac.authorization.k8s.io
-{{- end }}
--- a/templates/rbac.yaml
+++ b/templates/rbac.yaml
@@ -14,48 +14,6 @@ metadata:
     {{ $key }}: {{ $value | quote }}
   {{- end }}
 ---
-# The roles the cemanager needs cluster wide for
-# checking available resources.
-apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRole
-metadata:
-  name: coder-{{ .Release.Namespace }}-cluster
-rules:
-  - apiGroups:
-      - "" # indicates the core API group
-    resources:
-      # Lets us determine the cores/RAM available on a node
-      # to avoid letting a user select an unschedulable number.
-      - nodes
-      - pods
-      # Lets us determine what namespaces are available for
-      # scheduling. The permissions for a namespace are handled
-      # via regular namespaced Roles.
-      - namespaces
-    # We request read-only permissions at the cluster level to prevent
-    # the product from making any surprising mutations.
-    verbs:
-      - get
-      - list
----
-# Bind the cluster role to the coder serviceaccount.
-apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRoleBinding
-metadata:
-  namespace: {{ .Release.Namespace | quote }}
-  name: coder-{{ .Release.Namespace }}-clusterbinding
-subjects:
-  - kind: ServiceAccount
-    name: coder
-    namespace: {{ .Release.Namespace | quote }}
-roleRef:
-  kind: ClusterRole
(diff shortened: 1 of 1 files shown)


jawnsy (approved):
oooo

templates/rbac.yaml
```diff
@@ -68,17 +26,11 @@ rules:
...
       - pods
       # Required for creating registry pull secrets.
       - secrets
-      # Required for the code-server volume and the environments' home volumes.
+      # Required for the environments' home volumes.
       - persistentvolumeclaims
-      # Required for legacy reasons.
-      - statefulsets
-      # Every environment is a deployment.
-      - deployments
```
deansheather, line 81: We still have deployments code for migrating legacy deployments to pods, so this can't be removed. Pods are fairly new so I'm not comfortable removing that code yet

templates/rbac.yaml
```diff
@@ -68,17 +26,11 @@ rules:
...
-      # Every environment deployment has a service.
-      - services
       - pods
       # Required for creating registry pull secrets.
       - secrets
-      # Required for the code-server volume and the environments' home volumes.
+      # Required for the environments' home volumes.
       - persistentvolumeclaims
-      # Required for legacy reasons.
-      - statefulsets
```
deansheather, line 79: We have 2 kubemigrations that read statefulsets, so we need to delete those first

templates/rbac.yaml
```diff
@@ -68,17 +26,11 @@ rules:
       - "" # indicates the core API group
       - "apps"
     resources:
-      # Every environment deployment has a service.
-      - services
```
deansheather, line 72: We still use services, so this cannot be deleted, the wording could be changed to `Every workspace has a service` though.

templates/rbac.yaml
```diff
@@ -68,17 +26,11 @@ rules:
...
       - pods
       # Required for creating registry pull secrets.
       - secrets
-      # Required for the code-server volume and the environments' home volumes.
+      # Required for the environments' home volumes.
```
deansheather, line 32:
```suggestion
      # Required for workspace home volumes.
```

deansheather, line 32:
```suggestion
      # Required for workspace home volumes.
```

templates/rbac.yaml
```diff
@@ -102,12 +54,6 @@ rules:
       - watch
   - apiGroups:
       - "apps"
-    resources:
-      # Required for legacy reasons.
-      - statefulsets/scale
-      # Auto-off requires being able to update the
-      # replicas of a deployment from 1 to 0.
-      - deployments/scale
```
deansheather, line 110: Probably still need to keep this for the legacy deployment code

templates/rbac.yaml
```diff
@@ -14,48 +14,6 @@ metadata:
...
-    name: coder
-    namespace: {{ .Release.Namespace | quote }}
-roleRef:
-  kind: ClusterRole
-  name: coder-{{ .Release.Namespace }}-cluster
-  apiGroup: rbac.authorization.k8s.io
-
-{{- $namespaces := append .Values.namespaceWhitelist .Release.Namespace }}
-{{- range $namespaces }}
----
```
deansheather, line 58: You can't delete this yet unless you delete the deprecated `namespaceWhitelist` value. Deleting that value has the consequence that any old workspaces that haven't been migrated to workspace providers become inaccessible by coder. We probably need to add a big scary banner to the workspaces page for a few releases before doing that

deansheather (changes requested):

templates/rbac.yaml
kylecarbs: Ahh. Once the new network stuff comes to fruition I think we can actually remove this too :o

kylecarbs:
Too many issues to tackle in a near-term release. Better off punting this until 2.0.

PR #84, opened 2021-05-13
Make readOnlyRootFilesystem configurable

sreya:
This PR makes the `readOnlyRootFilesystem` field configurable via the `values.yaml`. It is enabled by default.

### DIFF
--- a/kube-linter.yaml
+++ b/kube-linter.yaml
@@ -11,10 +11,10 @@ checks:
     - no-readiness-probe
     - non-existent-service-account
     - privileged-container
+    - read-only-root-fs
     - run-as-non-root
     - ssh-port
     - writable-host-mount
-    - read-only-root-fs
   exclude:
     - no-extensions-v1beta
     - required-annotation-email
--- a/kube-linter.yaml
+++ b/kube-linter.yaml
@@ -8,10 +8,10 @@ checks:
     - mismatching-selector
     - no-anti-affinity
     - no-liveness-probe
+    - no-read-only-root-fs
     - no-readiness-probe
     - non-existent-service-account
     - privileged-container
-    - read-only-root-fs
     - run-as-non-root
     - ssh-port
     - writable-host-mount
--- a/kube-linter.yaml
+++ b/kube-linter.yaml
@@ -14,9 +14,9 @@ checks:
     - run-as-non-root
     - ssh-port
     - writable-host-mount
+    - read-only-root-fs
   exclude:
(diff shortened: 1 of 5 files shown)


templates/envproxy.yaml
```diff
@@ -70,7 +70,7 @@ spec:
 {{- end}}
           securityContext:
             allowPrivilegeEscalation: false
-            readOnlyRootFilesystem: true
+            readOnlyRootFilesystem: {{ .Values.envproxy.securityContext.readOnlyRootFilesystem }}
```
jawnsy, line 73: Can we do: `{{ .Values.envproxy.securityContext.readOnlyRootFilesystem | default true }}` here? This way users don't need to add it to their values file if they don't already have it in there https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-default-function

jawnsy (approved):

templates/cemanager.yaml
```diff
@@ -90,6 +90,9 @@ spec:
           ports:
             - name: tcp-cemanager
               containerPort: 8080
+          securityContext:
+            allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: {{ .Values.cemanager.securityContext.readOnlyRootFilesystem }}
```
jawnsy, line 95: With these changes I think we can move this linter setting from exclude to include: https://github.com/cdr/enterprise-helm/blob/766cc6d061e78c2c3005c0b52d65e8df5bdd2b27/kube-linter.yaml#L19

templates/envproxy.yaml
sreya: Is that not achieved by giving it a default value of `true` in the values.yaml?

jawnsy: I'm not sure actually, does it always apply the values.yaml included in the chart first, and then the user's values?

sreya: Yup, if you specify nothing in your "override" it'll take the values in the packaged `values.yaml`

jawnsy: cool stuff, thanks for the info

kube-linter.yaml
```diff
@@ -14,9 +14,9 @@ checks:
     - run-as-non-root
     - ssh-port
     - writable-host-mount
+    - read-only-root-fs
```
jawnsy, line 17: do you mind moving this between privileged-container and run-as-non-root? this list is alphabetically sorted

PR #85, opened 2021-05-13
feat: expose port 8443 on ce-manager for https

lilshoff:
Helm updates to support https://github.com/cdr/m/pull/8248

### DIFF
--- a/templates/cemanager.yaml
+++ b/templates/cemanager.yaml
@@ -177,6 +177,9 @@ spec:
     - name: tcp-cemanager
       port: 8080
       protocol: TCP
+    - name: tcp-cemanager-https
+      port: 8443
+      protocol: TCP
     - name: tcp-cemanager-turns
       port: 5349
       protocol: TCP


templates/cemanager.yaml
```diff
@@ -177,6 +177,9 @@ spec:
     - name: tcp-cemanager
       port: 8080
       protocol: TCP
+    - name: tcp-cemanager-https
```
jawnsy, line 180: does the ingress config need to be modified to forward traffic to this port, too?

jawnsy (approved):

templates/cemanager.yaml
lilshoff: Yeah so I'm in a bit of a weird situation rn with this...

This is to support https://github.com/cdr/m/pull/8248, which I guess is intended to only work after the ingress has been removed. 
So this helm change is only relevant for deployment configurations where ce-manager is directly accessible with an external IP and people want to access it over https without going through the ingress. 

TBH I'm a bit hesitant about both of these PRs because of their reliance on "future changes" 🤷 

Lmk what you think 🙏

deansheather: I think it's fine. We don't need to put this in the 1.20 changelog if we don't want customers to use it

deansheather (approved):

templates/cemanager.yaml
jawnsy: I agree with @deansheather, there's no harm in merging this, it just won't be useful unless people forward traffic to the pod through some means. If it makes your testing easier, then it makes sense to me to just merge this :)

Emyrk (approved):
I defer to the judgement of the others.

PR #86, opened 2021-05-20
fix: Update API versions of deprecated templates

kylecarbs:
During our deployment we had a few warnings of deprecation:

image

### DIFF
--- a/templates/ingress.yaml
+++ b/templates/ingress.yaml
@@ -57,7 +57,7 @@ metadata:
     app.kubernetes.io/name: ingress-nginx
     app.kubernetes.io/part-of: ingress-nginx
 ---
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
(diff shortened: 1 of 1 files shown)


deansheather (commented):
Does anything else need to be changed? It's kinda weird that nothing changes in our templates and we just have to change the version number.

We also need to make sure customers support these new versions first

jawnsy:
Yeah, the reason we haven't fixed these deprecated APIs is because we don't have a defined minimum Kubernetes support level, and we need to do some research to figure out what versions we need to support based on our customers. There was a card for this: https://app.clubhouse.io/coder/story/7570/switch-away-from-alpha-beta-api-versions-if-possible and I also started updating the docs (we currently say that we support 1.13 or later), but I think we need to do some research to figure out what our baseline version should be: https://github.com/cdr/docs/pull/232

We also use some of these APIs in our code (try.sh, some of our client-go code), so those can be cleaned up as well

kylecarbs:
Makes sense.

We'll be out of date with the latest Kube on August 22nd if we don't make these changes by then.

See: https://www.kubernetes.dev/resources/release/#tldr

jawnsy:
Superseded by #106

PR #89, opened 2021-05-22
chore: Add dashboard image back with cemanager replica

kylecarbs:
We aren't deprecating the dashboard until next release, so for now we can use a cemanager replica to essentially perform the same function.

### DIFF
--- a/templates/dashboard.yaml
+++ b/templates/dashboard.yaml
@@ -1,0 +1,91 @@
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    coder.deployment: dashboard
+  name: dashboard
+  namespace: {{ .Release.Namespace | quote }}
+  annotations:
+  {{- range $key, $value := .Values.deploymentAnnotations }}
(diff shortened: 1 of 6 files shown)


templates/dashboard.yaml
```diff
@@ -0,0 +1,91 @@
...
+    metadata:
+      labels:
+        coder.deployment: dashboard
+      annotations:
+      {{- range $key, $value := .Values.deploymentAnnotations }}
+        {{ $key }}: {{ $value | quote }}
+      {{- end }}
+    spec:
+      # coder:coder
+      securityContext:
```
jawnsy, line 32: Well, the linter works! :rotating_light: 

```
  /home/runner/work/enterprise-helm/enterprise-helm/build/kind/coder/templates/dashboard.yaml: (object: default/dashboard apps/v1, Kind=Deployment) container "dashboard" does not have a read-only root file system (check: no-read-only-root-fs, remediation: Set readOnlyRootFilesystem to true in your container's securityContext.)
```

jawnsy (approved):

PR #90, opened 2021-05-22
chore: Improve style of README

kylecarbs:
Prettied it up a bit to match code-server a bit more.

Also added a generate step in the CI... I wasn't even aware we generated the README.md... cool though 👍


.github/workflows/build.yml
```diff
@@ -31,6 +31,10 @@ jobs:
         if: always()
         run: ./scripts/fmt.sh
 
+      - name: Generate README.md
```
jawnsy, line 34: I think the fmt script already does this, because `make fmt` also re-generates the README.md based on the gotmpl https://github.com/cdr/enterprise-helm/blob/d51e11a277291eb192b707affd2befef51b3a289/scripts/fmt.sh#L18

It maybe isn't clear that's happening though, open to renaming `make fmt` to something else!

Makefile
```diff
@@ -38,6 +38,8 @@ fmt: README.md
 README.md: README.md.gotmpl values.yaml
 	@echo "--- Generating documentation"
 	helm-docs --template-files=$<
+	@echo -e "<!-- DO NOT EDIT. THIS IS GENERATED FROM README.md.gotmpl -->\n\n$$(cat README.md)" > README.md
```
jawnsy, line 41: Love this addition, thanks!

.github/workflows/build.yml
kylecarbs: Ahh... nah it was just my bad for not reading tbh. I'll remove the new script!

jawnsy: You can confirm this is happening by looking at the build output, btw:

image

Makefile
jawnsy: It doesn't really matter, but an alternative to doing this is just to add the text to the README.md.gotmpl itself at the top, and since it's a Go template, you can add a comment to explain that the line is for the generated output, like:

```
{{/* Add a comment to the generated output */}}
<!-- DO NOT EDIT. THIS IS GENERATED FROM README.md.gotmpl -->
```

But doing it in the makefile is also fine :man_shrugging:

.github/workflows/build.yml
jawnsy: We probably should have had separate `make fmt`/`make gen` steps, and set up `make fmt` to depend on make gen... I was just being lazy lol

README.md.gotmpl
```diff
@@ -1,39 +1,33 @@
-# Coder Helm Chart
+# Coder Helm Chart &middot; [![build](https://github.com/cdr/enterprise-helm/actions/workflows/build.yml/badge.svg)](https://github.com/cdr/enterprise-helm/actions/workflows/build.yml) [![Twitter Follow](https://img.shields.io/twitter/follow/CoderHQ?label=%40CoderHQ&style=social)](https://twitter.com/coderhq)
```
jawnsy, line 1: just a personal opinion, I think it's better to put these on their own line below the header line, because Markdown uses the header name to generate a link, and so the link to this header has some extraneous hyphens: `https://github.com/cdr/enterprise-helm/blob/34040562829bd72f9091bb955032ddd0ac4b89ee/README.md#coder-helm-chart---`

I also am not sure whether it matters, but I recently changed the monorepo README to refer to a specific event (`push`) and branch (`master`) in the markup: `[coder](https://github.com/cdr/m/actions/workflows/coder.yaml?query=branch%3Amaster)`

README.md.gotmpl
```diff
@@ -1,39 +1,33 @@
...
+
+```bash
+$ make README.md
+```
 
 ## Support
 
-If you have any questions or concerns about our Helm installer, use our [contact
-form](https://coder.com/contact) or send an email to support@coder.com.
+If you experience issues, have feedback, or want to ask a question, open an issue or
```
jawnsy, line 29: Why not the contact form? I think it goes to the same place, right?

README.md.gotmpl
```diff
@@ -1,39 +1,33 @@
...
-in the Helm values file (`values.yaml`), and we use `helm-docs` to generate the
-`README.md` file based on a template, `README.md.gotmpl`. When modifying content
-in the README or the `values.yaml` file, be sure to re-generate the README by
-running `make README.md`.
+Templates for Coder services live in the `templates` directory.
+Helm compiles templates with `values.yaml` when deploying.
+
+`README.md` is generated from `README.md.gotmpl` to ensure values are correct. Regenerate the readme:
+
+```bash
```
jawnsy, line 23: For shell commands, use `shell-session` because it highlights slightly differently (the $ indicating the prompt)

I also am not sure why we put Contributing/Support sections above the values listing, I imagine that most people that come to this page will just want to see the available values they can use...

jawnsy (approved):

README.md.gotmpl
kylecarbs: I'm good with that.

The default branch is used when one isn't specified, but I've updated the specific event!

README.md.gotmpl
kylecarbs: Agreed. Updated!

README.md.gotmpl
kylecarbs: Definitely agree. Going to move that section ordering.

PR #91, opened 2021-05-22
chore: Update chart branding

kylecarbs:
Our current chart is visible here: https://artifacthub.io/packages/helm/coder/coder

Inspired by: https://gitlab.com/gitlab-org/charts/gitlab/-/blob/master/Chart.yaml


jawnsy (approved):
Nice!

PR #92, opened 2021-05-23
fix: Move notes to templates/notes

kylecarbs:
Notes were never actually outputted before... now customers can detect deprecation notices when upgrading!

Run the following to test it out yourself:

```shell-session
$ helm install --dry-run --debug coder ./
```

image


jawnsy (approved):
Nice! There was a long standing card to fix this lol

jawnsy:
[ch1374]

PR #93, opened 2021-05-24
chore: Migrate cemanager to coderd

kylecarbs:
Enables migration from `cemanager` to `coderd` in values.yaml.

Deployments will output a deprecation notice to update the value.

Output when updating to 1.20 with built-in ingress:
image

Output when updating to 1.20 with custom ingress:
image

Allows customers to remove envproxy and the dashboard in 1.20, without a forced removal.

### DIFF
--- a/templates/NOTES.txt
+++ b/templates/NOTES.txt
@@ -7,38 +7,29 @@ Warning: `namespaceWhitelist` has been removed. Use multiple workspace providers
 {{- end }}
 
 {{- if .Values.dashboard }}
+Warning: The `dashboard` service has been deprecated.
 {{- if eq .Values.ingress.useDefault false }}
-Breaking
-{{- else -}}
-Warning
-{{- end -}}
-: The `dashboard` service will be removed in v1.21 (July 21st, 2021).
-{{- if eq .Values.ingress.useDefault false }}
-  All `dashboard` traffic must be routed to `cemanager` before v1.21.
+  All `dashboard` traffic can be routed to `cemanager`.
 {{- end }}
-  The dashboard is now served from `cemanager`. Delete `dashboard` in values.yaml.
-{{- end }}
+  Dashboard is now served from `cemanager`. Delete `dashboard` in values.yaml.
+{{ end }}
 
 {{- if .Values.envproxy }}
+Warning: The `envproxy` service has been deprecated.
 {{- if eq .Values.ingress.useDefault false }}
-Breaking
+  Coder now supports end-to-end encryption. To migrate, expose port 5349 to `cemanager`.
 {{- else }}
-Warning
-{{- end -}}
-: The `envproxy` service will be removed in v1.21 (July 21st, 2021).
-{{- if eq .Values.ingress.useDefault false }}
-  All `envproxy` traffic must be routed to `cemanager` before v1.21.
-{{- end }}
   Envproxy is now served from `cemanager`. Delete `envproxy` in values.yaml.
-{{- end }}
+{{- end}}
+{{ end }}
 
 {{- if .Values.podSecurityPolicyName }}
 Warning: `podSecurityPolicyName` is deprecated. It will be removed in v1.21 (July 21st, 2021).
 {{- end }}
 
 {{- if .Values.cemanager }}
-Breaking: The `cemanager` service will be renamed to `coderd` in v1.21 (July 21st, 2021).
+Warning: The `cemanager` service will be renamed to `coderd` in v1.21 (July 21st, 2021).
   The `cemanager` deployment, service, and ingress networking will reference `coderd`.
   If you do not reference `cemanager` externally, this change is not breaking.
   Rename `cemanager` to `coderd` in values.yaml to trigger this change.
-{{- end }}
+{{ end }}
--- a/templates/NOTES.txt
+++ b/templates/NOTES.txt
@@ -7,11 +7,29 @@ Warning: `namespaceWhitelist` has been removed. Use multiple workspace providers
 {{- end }}
 
 {{- if .Values.dashboard }}
-Warning: The `dashboard` service will be removed in v1.21 (July 21st, 2021).
+{{- if eq .Values.ingress.useDefault false }}
+Breaking
+{{- else -}}
+Warning
+{{- end -}}
+: The `dashboard` service will be removed in v1.21 (July 21st, 2021).
 {{- if eq .Values.ingress.useDefault false }}
   All `dashboard` traffic must be routed to `cemanager` before v1.21.
 {{- end }}
-  Delete `dashboard` in values.yaml to migrate.
+  The dashboard is now served from `cemanager`. Delete `dashboard` in values.yaml.
+{{- end }}
+
+{{- if .Values.envproxy }}
+{{- if eq .Values.ingress.useDefault false }}
+Breaking
+{{- else }}
+Warning
+{{- end -}}
+: The `envproxy` service will be removed in v1.21 (July 21st, 2021).
+{{- if eq .Values.ingress.useDefault false }}
+  All `envproxy` traffic must be routed to `cemanager` before v1.21.
+{{- end }}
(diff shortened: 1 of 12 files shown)


templates/coderd.yaml
```diff
@@ -128,7 +144,11 @@ spec:
             - name: STORAGE_CLASS
               value: {{ .Values.storageClassName | quote }}
             - name: TURN_ENABLED
+              {{- if hasKey .Values "cemanager" }}
               value: {{ .Values.cemanager.turn.enable | quote }}
+              {{- else }}
+              value: {{ .Values.coderd.turn.enable | quote }}
```
f0ssel, line 150: With `coderd`, TURN port shouldn't be optional IMO. If they move to the new pattern that also enables TURN by default I think

templates/dashboard.yaml
```diff
@@ -1,3 +1,6 @@
+# Optionally include the dashboard if it's specified.
+# The dashboard will be removed in v1.21.
```
f0ssel, line 2: We may need to keep the old kube Service around and just point it to coderd or something to help keep shit from breaking for people who don't heed the warnings. Idk if we should in this PR or just figure that out separately towards release

f0ssel (approved):

templates/coderd.yaml
kylecarbs: Good point!

templates/dashboard.yaml
kylecarbs: Agreed agreed.

If they exclude `dashboard` from the `values.yaml`, it just won't deploy it which is great!

templates/NOTES.txt
```diff
@@ -7,9 +7,29 @@ Warning: `namespaceWhitelist` has been removed. Use multiple workspace providers
 {{- end }}
 
 {{- if .Values.dashboard }}
-Warning: The `dashboard` service will be removed in v1.21 (July 21st, 2021). You can safely delete the `dashboard` property.
+Warning: The `dashboard` service has been deprecated.
```
jawnsy, line 10:
```suggestion
Warning: The `dashboard` service is deprecated since Coder v1.20, and will be removed in v1.21.
```

Not sure why the removal comment was changed, that seems useful?

templates/NOTES.txt
```diff
@@ -7,9 +7,29 @@ Warning: `namespaceWhitelist` has been removed. Use multiple workspace providers
...
 {{- if eq .Values.ingress.useDefault false }}
-
-Breaking: All `dashboard` traffic should be routed to `cemanager` before v1.21.
+  All `dashboard` traffic can be routed to `cemanager`.
 {{- end }}
+  Dashboard is now served from `cemanager`. Delete `dashboard` in values.yaml.
```
jawnsy, line 14:
```suggestion
  The `cemanager` serves the dashboard; delete `dashboard` in values.yaml.
```

templates/NOTES.txt
```diff
@@ -7,9 +7,29 @@ Warning: `namespaceWhitelist` has been removed. Use multiple workspace providers
...
+{{ end }}
+
+{{- if .Values.envproxy }}
+Warning: The `envproxy` service has been deprecated.
```
jawnsy, line 18:
```suggestion
Warning: The `envproxy` service is deprecated since Coder v1.20, and will be removed in v1.21.
```

templates/NOTES.txt
```diff
@@ -7,9 +7,29 @@ Warning: `namespaceWhitelist` has been removed. Use multiple workspace providers
...
+{{- if eq .Values.ingress.useDefault false }}
+  Coder now supports end-to-end encryption. To migrate, expose port 5349 to `cemanager`.
+{{- else }}
+  Envproxy is now served from `cemanager`. Delete `envproxy` in values.yaml.
+{{- end}}
+{{ end }}
+
+{{- if .Values.podSecurityPolicyName }}
+Warning: `podSecurityPolicyName` is deprecated. It will be removed in v1.21 (July 21st, 2021).
```
jawnsy, line 27: In general, deprecations should come with a suggestion for what to do to migrate, we shouldn't remove things unless there is an alternative?

```suggestion
Warning: `podSecurityPolicyName` is deprecated since Coder v1.20 and will be removed in v1.21.
```

templates/coderd.yaml
```diff
@@ -82,18 +90,26 @@ spec:
...
+        - name: {{ include "coder.serviceName" . }}
+          {{- if hasKey .Values "cemanager" }}
           image: {{ .Values.cemanager.image | quote }}
+          {{- else }}
+          image: {{ .Values.coderd.image | quote }}
+          {{- end }}
           imagePullPolicy: {{ .Values.imagePullPolicy | quote }}
           ports:
-            - name: tcp-cemanager
+            - name: tcp-{{ include "coder.serviceName" . }}
```
jawnsy, line 103: do we need a port here for 8443? how about 5349?

values.yaml
```diff
@@ -206,18 +202,19 @@ dashboard:
       cpu: "250m"
       memory: "512Mi"
 
+# Deprecated: Being removed in v1.21.
```
jawnsy, line 205:
```suggestion
# envproxy is deprecated as of v1.20 and will be removed in v1.21, as it unnecessary with the Networking V2.
```

jawnsy (approved):

templates/NOTES.txt
```diff
@@ -7,9 +7,29 @@ Warning: `namespaceWhitelist` has been removed. Use multiple workspace providers
...
-Breaking: All `dashboard` traffic should be routed to `cemanager` before v1.21.
+  All `dashboard` traffic can be routed to `cemanager`.
 {{- end }}
+  Dashboard is now served from `cemanager`. Delete `dashboard` in values.yaml.
+{{ end }}
+
+{{- if .Values.envproxy }}
+Warning: The `envproxy` service has been deprecated.
+{{- if eq .Values.ingress.useDefault false }}
+  Coder now supports end-to-end encryption. To migrate, expose port 5349 to `cemanager`.
```
deansheather, line 20:
```suggestion
  Coder now supports end-to-end encryption. To migrate, forward TCP 5349 traffic to `cemanager`.
```

templates/coderd.yaml
```diff
@@ -50,7 +54,11 @@ spec:
 {{- include "coder.serviceTolerations" . | indent 6 }}
       initContainers:
         - name: migrations
+          {{- if hasKey .Values "cemanager" }}
           image: {{ .Values.cemanager.image | quote }}
+          {{- else }}
+          image: {{ .Values.coderd.image | quote }}
+          {{- end }}
```
deansheather, line 61: This is very messy since we have a lot of values. I think we should just force people to change `cemanager` to `coderd` in this version instead since it's a one line change.

values.yaml
deansheather:
```suggestion
# Deprecated: envproxy will be removed in v1.21 as it will be superseded by Networking V2.
```

deansheather (changes requested):
I don't think letting customers wait one release before changing one line from `cemanager` to `coderd` in their values file is worth putting an `if/else` everywhere we reference a `coderd` value. We should just force customers to change the line in 1.20

templates/coderd.yaml
kylecarbs: I think we need at least 1 version of deprecation before forcing a switch.

Not because changing the value is hard, but customers may have external services that reference `cemanager` for logs and monitoring... us straight-up breaking that without warning sounds sus.

templates/coderd.yaml
kylecarbs: I've added 5349. 8443 is related to the TLS work, so I'd rather that be a separate PR.

templates/coderd.yaml
deansheather: In that case we can keep the service and resource names for one version then and change them in the next version

deansheather (approved):

PR #95, opened 2021-05-27
 [ch13351] Remove environments sa

f0ssel:
Removes the `environments` sa, and will use the `coder` sa for workspaces. This is fine because we don't mount credentials, so it's just used to apply any specified psp.


templates/_environments.tpl
```diff
@@ -1,37 +1,3 @@
...
-  namespace: {{ . | quote }}
-  name: environments
-{{- end }}
-{{- end }}
-{{- end }}
-{{/* 
-  coder.environments.configMap defines configuration that is applied
-  to user environments.
-*/}}
-{{- define "coder.environments.configMap" }}
```
f0ssel, line 23: We stopped using this a while back in favor of the `coder.environments.configMapEnv` template instead

templates/psp.yaml
```diff
@@ -33,7 +33,7 @@ subjects:
     name: envproxy
     namespace: {{ .Release.Namespace | quote }}
   - kind: ServiceAccount
-    name: environments
+    name: coder
```
jawnsy, line 36: apologies if I'm missing something, just for my edification here:

1. what's envproxy using this service account for?
2. in a coderd world, are we going to use this coder account, or are we going to use `cemanager`?
3. are we going to have a `coder` service account as well as `coderd` (formerly `cemanager`) service account?

f0ssel:
Closing since this change would break `namespaceWhitelist` which we aren't ready to remove quite yet

PR #96, opened 2021-05-31
fix: cemanager migration error if accessURL isn't set

kylecarbs:


jawnsy (approved):
would be nice if there was a way to test things like this in CI... I started adding some example values files in `examples/` and at one point @deansheather was working on using the Helm Go APIs to write tests

templates/dashboard.yaml
```diff
@@ -52,7 +52,9 @@ spec:
               containerPort: 3000
           securityContext:
             allowPrivilegeEscalation: false
+            {{- if hasKey .Values.dashboard "securityContext" }}
```
deansheather, line 55: Is this necessary? This value has a default from values.yaml.

values.yaml
```diff
@@ -133,7 +133,7 @@ imagePullPolicy: Always
 
 # Contains configuration the REST API handling CRUD operations to
 # the platform.
-coderd:
+cemanager:
```
deansheather, line 136: If you're changing this you should change all of the comments below

deansheather (approved):

PR #99, opened 2021-06-09
remove note about envproxy deprecation

f0ssel:
Since this is being delayed to a patch release


kylecarbs (approved):
Should we promote the dashboard deletion in this release? I guess it's fine. Hard for me to tell what message I put there on mobile, but feel free to adjust if it's not clear.

f0ssel:
Yeah it's not related and can be deprecated still

PR #102, opened 2021-06-10
Fix regression so that SSL cert secrets are mounted in properly

nathanpotter:


nathanpotter:
I agree, that would be really nice, I think it's definitely something we can do, we just need to build the infrastructure around it.

kylecarbs (approved):

PR #106, opened 2021-06-16
feat: remove deprecated v1beta APIs

jawnsy:
Replace the use of deprecated apiVersions with the corresponding
stable versions.

--

Tested this manually in the latest monorepo.

### DIFF
--- a/templates/ingress.yaml
+++ b/templates/ingress.yaml
@@ -59,7 +59,7 @@ metadata:
     app.kubernetes.io/name: ingress-nginx
     app.kubernetes.io/part-of: ingress-nginx
 ---
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: nginx-ingress-clusterrole-{{ .Release.Namespace }}
@@ -125,7 +125,7 @@ rules:
       - list
       - watch
 ---
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
   name: nginx-ingress-role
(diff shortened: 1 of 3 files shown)


deansheather (approved):

kylecarbs (approved):

templates/ingress.yaml
```diff
@@ -370,86 +370,126 @@ metadata:
...
-          servicePort: 8080
+          service:
+            name: {{ include "coder.serviceName" . }}
+            port:
+              number: 8080
           {{- end }}
     {{- end }}
     {{- else }}
-      - path: /proxy/*
+      - path: /proxy
```
coadler, line 441: This seems to obviate the `usePathWildcards` option. Do we no longer need them? Last time I played with this, `gce-ingress` required wildcards, but maybe this is fixed now by specifying `pathType: Prefix`. If so we should just remove that option.

coadler (approved):

templates/ingress.yaml
jawnsy: I did test with my local dev deploy but am not sure how to test otherwise since I'm not familiar with what the problem was

coadler: I'll test it out

jawnsy: @coadler What was the outcome of your testing? Should we remove that option?

templates/ingress.yaml
```diff
@@ -372,88 +372,73 @@ metadata:
...
-          serviceName: {{ include "coder.serviceName" . }}
-          servicePort: 8080
+          service:
+            name: {{ include "coder.serviceName" . }}
+            port:
+              number: 8080
           {{- end }}
-    {{- end }}
-    {{- else }}
-      - path: /proxy/*
```
jawnsy, line 422: @coadler Removed all this stuff, `pathType: Prefix` should cover these, I think. I tested with my dev deploy but that's it so far.

jawnsy:
Gonna merge this, if we have a problem with the ingress and path wildcards, we can always revert or look into it more closely...

PR #107, opened 2021-06-17
feat: Refactor Helm values for new architecture changes

kylecarbs:
⚠ So many people are tagged because these are quite opinionated and difficult to reverse decisions. Please review carefully!

Refactors `values.yaml` for easier deployment. This does NOT address any actual code-changes, and is not ready to merge as-is. This is to collect feedback on this state before implementation.

Backwards compatibility will be supported on all fields.

Will do in future PRs:
- Add `certs.*` support.
- Output transition notice for `environments.*`.
- Explain rationale for rename, and simplification with docs.

List of changes:
- ↪ Move `storageClassName` to `postgres.default.storageClassName`
  We only persist data for the `postgres` anyways with this field.
  Storage class is now a Provider option.
- ↪ Move `serviceType` to `services.type`.
- ⚠ Deprecate `podSecurityPolicyName`.
  It's been deprecated for a while, but hasn't been outputting in NOTES.txt.
- ↪ Move `clusterDomainSuffix` to `services.clusterDomainSuffix`
  ~Nobody seems to be aware why this exists. It doesn't exist in GitLab's chart either.~
  Easier to reference all `service` communication inside this object.
- ↪ Move `ingress.useDefault` to `ingress.enable`.
  We'll use `ingress.useDefault` for backwards compatibility, but we shouldn't be deploying an Ingress controller.
- ⚠ Deprecate `ingress.service.annotations`.
  We can use `services.annotations` instead.
- ↪ Move `ingress.service.externalTrafficPolicy` to `ingress.externalTrafficPolicy`.
- ⚠ Deprecate `ingress.deployment`.
  We'll be removing the ingress controller, so there won't be a deployment here anymore.
- ⚠ Deprecate `ingress.additionalAnnotations`.
  It's been deprecated for a while, but hasn't been outputting in NOTES.txt.
- 👋 Maybe deprecate `ingress.podSecurityPolicyName`?
  Help please, not certain why we'd need this.
- ⚠ Deprecate `ingress.host`.
  This will be a configurable option in the UI via LetsEncrypt.
- ⚠ Deprecate `ingress.useDefault`.
  It's easier to process `enable` then `useDefault` and `enable`. We're dropping the Ingress object in favor of a simpler LoadBalancer approach. 
- ⚠ Deprecate `devurls.host`.
  This will be a configurable option in the UI via LetsEncrypt.
- ⚠ Deprecate `ingress.usePathWildcards`.
  Doesn't seem applicable if we remove the Ingress type.
- ⚠ Deprecate `ingress.annotations`.
  We can use `services.annotations` instead.
- ⚠ Deprecate `ingress.tls.enable`.
  TLS will be toggled via the UI.
- ⚠ Deprecate `imagePullPolicy`.
  Seems like this being configurable is likely to cause more harm than good.
- ↪ Move `postgres.useDefault` to `postgres.default.enable`.
  There were values nested under a separate `timescale` field that wasn't coherent.
- ↪ Move `timescale.image` to `postgres.default.image`.
- ↪ Move `timescale.resources` to `postgres.default.resources`.
- ⚠ Deprecate `certs`.
  Not much [history for this addition](https://github.com/cdr/enterprise-helm/commit/ada298f51539b524921dd9507b3cb3973b125295)... we haven't documented this value very well, so I'm not sure of usage.
- ⚠ Deprecate `namespaceWhitelist`.
  It's been deprecated for a while, but hasn't been outputting in NOTES.txt.
- ⚠ Deprecate `ssh`.
  Moved to a [Provider configuration option](https://app.clubhouse.io/coder/story/11821/change-enablessh-global-setting-to-enableexternalconnect-setting-per-workspace-provider).
- ⚠ Deprecate `environments`.
  Moved to a [Provider configuration option](https://app.clubhouse.io/coder/story/14031/move-toleration-and-nodeselector-options-to-workspace-provider-configuration).
- ↪ Move `serviceTolerations` to `services.tolerations`.
- ↪ Move `deploymentAnnotations` to `services.annotations`.
- ⚠ Deprecate `serviceAccount`.
  Dependent on whether we want to ship a built-in provider. Not an immediate requirement, and can totally be dropped.
- ↪ Move `ingress.tls` to `coderd.tls`.
  TLS will be managed by coderd directly.
- ↪ Move `ingress.*` to `coderd.serviceSpec.*`.
  We already ship a coderd service that uses a simpler `serviceType`. By default, this is configured to be exposed. It can be changed to be a `ClusterIP` to be made internal, or really whatever a customer would want.

## Ingress Configuration

We currently ship an nginx Ingress Controller as we needed to route traffic among multiple services. With Networking v2, we no longer need to do this.

A simple Service with type LoadBalancer can route all traffic to the application. If customers would like to customize their ingress to MITM, drop TLS, or anything else, they totally can! We'll document the required ports to be exposed very well.


templates/dashboard.yaml
```diff
@@ -1,5 +1,6 @@
 # Optionally include the dashboard if it's specified.
 {{- if .Values.dashboard }}
+{{- if .Values.dashboard.image }}
```
f0ssel: I really like this way of rolling out removals

values.yaml
```diff
@@ -1,402 +1,144 @@
...
+  # coderd.image -- Injected by Coder during release.
+  image: ""
+  # coderd.replicas -- The number of Kubernetes Pod replicas.
+  replicas: 1
+  # coderd.replica -- Deploy a replica to geodistribute access to
+  # workspaces for lower latency.
+  replica:
+    # coderd.replica.enable -- Run coderd as a replica pointing to a primary deployment.
+    enable: false
+    # coderd.replica.primaryURL -- URL of the primary deployment.
```
f0ssel, line 12: an example of url format like `https://coder.example.com` may help here

values.yaml
```diff
@@ -1,402 +1,144 @@
...
+        cpu: "250m"
+        memory: "1Gi"
+        # postgres.default.resources.requests.storage -- Specifies the size of the volume claim
+        # for persisting the database.
+        storage: "10Gi"
+      limits:
+        cpu: "250m"
+        memory: "1Gi"
+
+# services -- Kubernetes Service configuration that applies to Coder services.
```
f0ssel, line 106: Can we roll this into coderd now that it's the only service we run?

values.yaml
```diff
@@ -1,402 +1,144 @@
...
-# resolution to work correctly.
-clusterDomainSuffix: ".svc.cluster.local"
+# envbox -- Required for running Docker inside containers. See requirements:
+# https://coder.com/docs/coder/v1.19/admin/workspace-management/cvms
+envbox:
+  # envbox.image -- Injected by Coder during release.
+  image: ""
 
-# contains configuration for the bundled ingress controller.
+# ingress -- Configure access to your deployment.
```
f0ssel, line 37: I don't think we should use the ingress term here unless we are configuring an actual Ingress. For kube Service config, we should have a key like `coderd.serviceType: LoadBalancer` or `coderd.serviceType: ClusterIP`

values.yaml
kylecarbs: I was thinking that too, but we still optionally run PostgreSQL.

values.yaml
kylecarbs: I'd say that's reasonable.

values.yaml
f0ssel: ah yeah makes sense to apply to that too

values.yaml
```diff
@@ -1,402 +1,144 @@
...
-  additionalAnnotations: []
-  # ingress.annotations -- Additional annotations to be used when creating the
-  # ingress. These only apply to the Ingress Kubernetes kind. The annotations
-  # can be used to specify certificate issuers or other cloud provider-specific
-  # integrations.
-  annotations: {}
   # ingress.tls -- TLS options for the ingress. The hosts used for the tls
   # configuration come from the ingress.host and the devurls.host variables. If
   # those don't exist, then the TLS configuration will be ignored.
   tls:
```
f0ssel, line 54: I'd ideally see this as a `coderd.tls` so people don't think it's part of some ingress tls kube resource, since this is only for internal use in the app.

kylecarbs: That's much more reasonable. I agree, will move!

kylecarbs: Done!

values.yaml
kylecarbs: I put these all under a `serviceSpec` variable that just gets injected. Kinda a catch-all.

values.yaml
```diff
@@ -1,402 +1,134 @@
...
+  # coderd.replicas -- The number of Kubernetes Pod replicas.
+  replicas: 1
+  # coderd.serviceSpec -- Specification to inject for the coderd service. See:
+  # https://kubernetes.io/docs/concepts/services-networking/service/
+  serviceSpec:
+    type: LoadBalancer
+    loadBalancerIP: ""
+    loadBalancerSourceRanges: []
+  # coderd.tls -- TLS configuration for coderd.
+  # These options will override dashboard configuration.
```
f0ssel, line 14: You can't do any TLS config in the dash today, so maybe we hold off on this comment?

kylecarbs: I'd call this an end-state for 1.21. Since @lilshoff already did all the API stuff here, I'm confident we could do the UI by 1.21.

values.yaml
```diff
@@ -1,402 +1,134 @@
...
+  passwordSecret: ""
+  # postgres.sslMode -- Provides variable levels of protection for
+  # the PostgreSQL connection. For acceptable values, see:
+  #  https://www.postgresql.org/docs/9.1/libpq-ssl.html
+  sslMode: "require"
 
-# Services contains global configuration that applies to all Coder managed
-# services.
+  # postgres.default -- Configure a built-in PostgreSQL deployment.
+  default:
```
coadler, line 76: If we're gonna break this backwards compat, I think we should move to shipping the official Postgres helm as a subchart.

values.yaml
```diff
@@ -1,402 +1,134 @@
...
+  # coderd.image -- Injected by Coder during release.
+  image: ""
+  # coderd.replicas -- The number of Kubernetes Pod replicas.
+  replicas: 1
+  # coderd.serviceSpec -- Specification to inject for the coderd service. See:
+  # https://kubernetes.io/docs/concepts/services-networking/service/
+  serviceSpec:
+    type: LoadBalancer
+    loadBalancerIP: ""
+    loadBalancerSourceRanges: []
```
coadler, line 12: Seem to have lost the docs on these

coadler (approved):

values.yaml
f0ssel: Or just give a basic tutorial on how to do it in the docs with the official chart, and we ship nothing?

kylecarbs: This wouldn't break backwards compat as it is right now though.

deansheather:
We can't remove clusterDomainSuffix, customers have custom suffixes and it's necessary for timescale and envproxy (for now).

values.yaml
coadler: Why wouldn't it? It seems like we're moving the resources and stuff from `timescale` to `postgres.default`.

values.yaml
```diff
@@ -1,402 +1,134 @@
...
-
-# serviceType -- See the following for the different serviceType options and
-# their use:
-# https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types
-serviceType: "ClusterIP"
-
-# podSecurityPolicyName -- The name of the pod security policy to apply to all
-# Coder services and user environments. The optional ingress has its own field
-# for pod security policy as well.
-podSecurityPolicyName: ""
```
jawnsy, line 16: [PodSecurityPolicy has been deprecated](https://kubernetes.io/blog/2021/04/06/podsecuritypolicy-deprecation-past-present-and-future/) and [should be replaced with custom admission controllers](https://www.antitree.com/2020/11/pod-security-policies-are-being-deprecated-in-kubernetes/), so I think it's fine to remove the annotation from our charts. We might want to contact the affected customers to inform them of the change, though.

values.yaml
```diff
@@ -1,402 +1,134 @@
...
-  useDefault: true
-  # ingress.usePathWildcards -- Whether or not the ingress object should use
-  # path wildcards, i.e., ending with "/*". Some ingresses require this
-  # while others do not. You should check which path style your ingress
-  # requires. For ingress-nginx this should be set to false.
-  usePathWildcards: false
-  # ingress.host -- The hostname to use for accessing the platform. This can
-  # be left blank, and the user can still access the platform from the external
-  # IP or a DNS name that resolves to the external IP address.
-  host: ""
```
jawnsy, line 42: How does someone configure `ingress.host` for the first time - don't we have a chicken-and-egg situation since we need a host to set up TLS, and we need TLS in order to log in to the admin console securely?

values.yaml
```diff
@@ -1,402 +1,134 @@
...
-  # string if the database does not require a password to connect.
-  passwordSecret: ""
-  # postgres.sslMode -- Determines how the connection is made to the database.
-  # The acceptable values are: `disable`, `allow`, `prefer`, `require`,
-  # `verify-ca`, and `verify-full`.
-  sslMode: "require"
-
-# imagePullPolicy -- Sets the policy for pulling a container image across all
-# services.
-imagePullPolicy: Always
```
jawnsy, line 132: Removing configuration of imagePullPolicy is okay; it can be useful to have this configurable because setting it to Always is required if enforcing image privacy for private repositories (the authorization check happens on image pulls, so using IfNotPresent means that someone can start a container with an image that they shouldn't have access to, in case the cache is warm).

I've heard the argument that people that are concerned about this would enforce it with a web hook, which seems reasonable to me

values.yaml
```diff
@@ -1,402 +1,134 @@
...
     token: ""
-    # logging.splunk_channel -- Optional. Specify the channel that you'd
-    # like to associate with all messages.
+    # logging.splunk_channel -- Optional. Specify the channel
+    # to associate messages with.
     channel: ""
-
-# serviceAccount configures the service account used for all Coder deployed
-# services.
-serviceAccount:
```
jawnsy, line 385: Don't we still need a serviceAccount to run coderd itself? What service account do we use for this? Is this for control plane services such as coder-service, or are these used for workspaces?

values.yaml
```diff
@@ -1,402 +1,134 @@
...
-  # 
-  # eg.
-  # nodeSelector:
-  #   disktype: ssd
-  nodeSelector: {}
-
-# certs -- Describes CAs that should be added to Coder services. These certs
-# are NOT added to environments.
-certs:
-  secret:
```
jawnsy, line 313: I think it's useful to have a way to inject certs into our pods; we should expand the feature rather than deprecating it, IMO. The new TLS support is great for handling server certs (public/private keys) but that doesn't cover outbound connections from Coder to other services:

1. CA certificates that Coder uses to validate the servers to which it connects
2. Client certificates that Coder presents to remote servers

values.yaml
jawnsy: I don't think timescale nor the postgresql images on Docker Hub support running in OpenShift, primarily because of the runAsUser uid.

CrunchyData is a company that provides a postgresql image for OpenShift, so if we do subchart stuff, that should be something that we consider.

values.yaml
kylecarbs: This should have an RFC, but I propose we remove the “built-in” provider in entirety.

values.yaml
coadler: > I don't think timescale nor the postgresql images on Docker Hub support running in OpenShift, primarily because of the runAsUser uid.
> 
> CrunchyData is a company that provides a postgresql image for OpenShift, so if we do subchart stuff, that should be something that we consider.

I think the benefit of just using the public chart is they could easily replace the image if necessary.

values.yaml
jawnsy: I think that all pods are associated with a service account. If we don't provide a service account name, then it will use the [`default` service account in the namespace](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/):

> A service account provides an identity for processes that run in a Pod.

and

> When you create a pod, if you do not specify a service account, it is automatically assigned the `default` service account in the same namespace.

This is true whether or not we mount the service account token into our pod (controlled with `automountServiceAccountToken: false` in the service account or Pod spec). The service account is associated with pull secrets, which we might need if the coderd image is in an air-gapped registry that requires password authentication.

Kube-Linter will emit a warning if pods use the default: https://docs.kubelinter.io/#/generated/checks?id=default-service-account - I'm not clear why this is a bad practice, but it seems to be :man_shrugging:

values.yaml
```diff
@@ -1,402 +1,134 @@
...
-  image: ""
-  # envproxy.securityContext -- Contains fields related to the envproxy
-  # container's security context (as opposed to the pod).
+  # coderd.replica -- Deploy a replica to geodistribute access to
+  # workspaces for lower latency.
+  replica:
+    # coderd.replica.enable -- Run coderd as a replica pointing to a primary deployment.
+    enable: false
+    # coderd.replica.primaryURL -- URL of the primary deployment.
+    # eg. us-east.coder.myorg.com
```
deansheather, line 26:
```suggestion
    # eg. https://coder.myorg.com
```

values.yaml
```diff
@@ -1,402 +1,134 @@
...
-  # envproxy.replicas -- The number of replicas to run of the envproxy.
-  replicas: 1
-  # envproxy.image -- Injected during releases.
-  image: ""
-  # envproxy.securityContext -- Contains fields related to the envproxy
-  # container's security context (as opposed to the pod).
+  # coderd.replica -- Deploy a replica to geodistribute access to
+  # workspaces for lower latency.
+  replica:
+    # coderd.replica.enable -- Run coderd as a replica pointing to a primary deployment.
```
deansheather, line 23: Make a note to link to docs so people know what this means

values.yaml
```diff
@@ -1,402 +1,134 @@
...
-
-# podSecurityPolicyName -- The name of the pod security policy to apply to all
-# Coder services and user environments. The optional ingress has its own field
-# for pod security policy as well.
-podSecurityPolicyName: ""
-
-# clusterDomainSuffix -- If you've set a custom default domain for your
-# cluster, you may need to remove or change this DNS suffix for service
-# resolution to work correctly.
-clusterDomainSuffix: ".svc.cluster.local"
```
deansheather, line 21: Can't go if we support bundled postgres. Multiple customers have custom suffixes.

values.yaml
deansheather: I'm in favor of getting rid of timescale in favor of postgres subchart or nothing entirely.

values.yaml
deansheather: Is it that unreasonable for people to log into the admin panel over HTTP during setup? If they really don't want to they can setup their own certificates and set the secret names on coderd.tls

values.yaml
deansheather: I doubt we'll ever remove the built-in provider, we'll probably just migrate it to the new provider format by orphaning the resources or something.

Also, AFAIK default service account with `automountBlah: false` is fine.

values.yaml
jawnsy: Yeah, that's a fair point @deansheather, I agree with that

values.yaml
kylecarbs: @coadler we can alias them internally in the chart to make transition simple. I mean that it wouldn't create a "new" Postgres instance.

Using another chart probably isn't something I'll do here.

values.yaml
kylecarbs: Can we have this set to "Always", always?

values.yaml
kylecarbs: Do we need this option? Bit confused on this.

values.yaml
kylecarbs: Could you write up the syntax of what you imagine? I'm happy to implement, just not as familiar with naming conventions in that realm.

values.yaml
jawnsy: I think that setting to `Always` means ignoring cache, so it seems like a performance/security tradeoff to me? Maybe we can test it out some more, I'm not sure if `Always` just means it'll check if the image is up-to-date and use the cache if so, or if it would re-download all the image layers

kylecarbs: `Always` ensures images exist locally with hashes specified in the image manifest.

values.yaml
jawnsy: Well, if you leave out the option then you'll end up with a service account mount (including token) for the default service account, which isn't authorized to do anything by default. I think it's best to set `automountServiceAccountToken: false`, yes

values.yaml
jawnsy: Ah, TIL that `Always` will still use image cache, that makes sense!

It's not very well-linked in the docs, the [page that describes images](https://kubernetes.io/docs/concepts/containers/images/#updating-images) does not link to the [page that describes the details of how imagePullPolicy works](https://kubernetes.io/docs/concepts/configuration/overview/#container-images), though the latter links to the former. Good to know!

> `imagePullPolicy: Always`: every time the kubelet launches a container, the kubelet queries the container image registry to resolve the name to an image digest. If the kubelet has a container image with that exact digest cached locally, the kubelet uses its cached image; otherwise, the kubelet downloads (pulls) the image with the resolved digest, and uses that image to launch the container.

templates/rbac.yaml
```diff
@@ -14,102 +14,30 @@ metadata:
...
-  - apiGroups:
-      - storage.k8s.io
-    resources:
-      - storageclasses
-    verbs:
-      - get
-      - list
-      - watch
+  - apiGroups: ["", "apps", "networking.k8s.io"] # "" indicates the core API group
+    resources: ["persistentvolumeclaims", "pods", "deployments", "services", "secrets", "pods/exec","pods/log", "events", "networkpolicies"]
```
jawnsy, line 25: the previous list seems easier to read (especially when the list changes, and since we can specify comments per entry). this one lists a lot of required rules but doesn't indicate why Coder needs them

values.yaml
jawnsy: just leave these out for now - from the directory structure, it looks like the `certs/` directory was meant to hold cert files, which we could load into Secrets using https://helm.sh/docs/chart_template_guide/accessing_files/#configmap-and-secrets-utility-functions

so I think having a way to specify filenames would be convenient, as well as to handle all the different types of certs that people would want to configure, as defined in the RFC: https://www.notion.so/TLS-certificates-e17259ebcaf84f64969ba6a9351d38a7

kylecarbs: Going to leave this in for right now and just move it. No need for action on your part!

templates/coderd.yaml
```diff
@@ -198,6 +198,34 @@ data:
...
+  namespace: {{ .Release.Namespace | quote }}
+spec:
+{{ toYaml .Values.coderd.serviceSpec | indent 2 }}
+  selector:
+    coder.deployment: {{ include "coder.serviceName" . }}
+  ports:
+    - name: tcp-{{ include "coder.serviceName" . }}
+      port: 80
+      targetPort: 8080
+      protocol: TCP
```
jawnsy, line 220: while you're in here, do you also want to add an `appProtocol`? https://kubernetes.io/docs/concepts/services-networking/service/#application-protocol

This was introduced as beta in 1.19, so I think should be OK given our version support policy https://v1-19.docs.kubernetes.io/docs/concepts/services-networking/service/#application-protocol

templates/coderd.yaml
```diff
@@ -194,46 +179,74 @@ spec:
...
   ports:
     - name: tcp-{{ include "coder.serviceName" . }}
-      port: 8080
+      port: 80
+      targetPort: 8080
       protocol: TCP
+      appProtocol: http
     - name: tcp-{{ include "coder.serviceName" . }}-https
-      port: 8443
+      port: 443
```
jawnsy, line 223: Nice! This means we can use privileged ports without running as root :tada: (This was a problem for some customers with nginx-ingress)

templates/rbac.yaml
kylecarbs: It's the exact same as what's in our UI when adding a new provider. We _can_ make that different, but I felt it led to confusion.

values.yaml
```diff
@@ -1,402 +1,160 @@
...
+    # coderd.serviceSpec.loadBalancerIP -- Set the external IP address of the
+    # Ingress service.
+    loadBalancerIP: ""
+    # coderd.serviceSpec.loadBalancerSourceRanges -- Traffic through the LoadBalancer
+    # will be restricted to the specified client IPs. This field will be ignored if
+    # the cloud provider does not support this feature.
+    loadBalancerSourceRanges: []
+  # coderd.tls -- TLS configuration for coderd.
+  # These options will override dashboard configuration.
   tls:
```
jawnsy, line 25: @kylecarbs What do you think of this?

The goal is to make it easy for people to configure any number of certificates
and keys, which can be used for different purposes:

1. Server certificate to offer to users
2. CA certificate to validate external services
3. Client certificate to offer as authentication to external services (mTLS)

The default secretNames will be used as CA certs (scanned using the certificate
handling code, since we look recursively for these files)

```yaml
tls:
  certificates:
    # A list of secret names. The contents of this secret would be mounted
    # to /etc/ssl/certs with the same name (e.g., /etc/ssl/certs/ca-certs)
    secretNames:
      - ca-certs
      - other-certs
    # A list of files stored under ./ssl/certs. The contents of these files
    # would be added as a secret that Coder creates. We can use Files.Glob
    # to load the files into a Secret or ConfigMap: https://helm.sh/docs/chart_template_guide/accessing_files/#configmap-and-secrets-utility-functions
    # Maybe ConfigMap makes sense since certificates are public keys, not
    # really secret.
    files:
      - ca.crt
      - server.crt
  keys:
    # A list of secret names. The contents of this secret would be mounted
    # to /etc/ssl/private with the same name (e.g., /etc/ssl/private/ca-certs)
    secretNames:
      - server-keys
      - client-keys
    # A list of files stored under ./ssl/private. The contents of these files
    # would be added as a secret that Coder creates. We can use Files.Glob
    # to load the files into a Secret or ConfigMap: https://helm.sh/docs/chart_template_guide/accessing_files/#configmap-and-secrets-utility-functions
    files:
      - server.key
      - client.key
```

We would still need a separate mechanism to change the SSL_CLIENT_KEY_FILE etc variables, though.

kylecarbs: Would it be correct to say:

- `certificates` are loaded as CAs
- `keys` are client + server keys? (bit confusing to me here)

jawnsy: certificates can be the certs for your keypairs too! like:

- ca-certs.crt -> bundle of certs
- server.crt -> certificate corresponding to the server cert (presented to users connecting to Coder)
- client.crt -> certificate corresponding to the client cert (presented to servers Coder is connecting to)

keys are the private keys corresponding to our certs (server or client), e.g.:

- server.key
- client.key

kylecarbs: Ahh I believe I understand.

I'm going to add these in another PR. I'm worried about scope-creep getting the best of us here.

values.yaml
```diff
@@ -1,397 +1,165 @@
...
+  default:
+    # postgres.default.enable -- Deploys a PostgreSQL instance. We recommend
+    # using an external PostgreSQL instance in production.
+    # If true, all other values are ignored.
+    enable: true
+    # postgres.default.image -- Injected by Coder during release.
+    image: ""
+    # postgres.default.storageClassName -- Set the storageClass to store
+    # the database.
+    storageClassName: ""
```
ericpaulsen, line 110: Just want to clarify my understanding from the PR description above - since `StorageClassName` has been moved under `postgres`, that means I would need to set my PVC StorageClass at the workspace provider level, correct?

values.yaml
jawnsy: Yeah totally! I agree, a separate PR is better

values.yaml
```diff
@@ -1,397 +1,165 @@
...
+        memory: "1Gi"
+
+# services -- Kubernetes Service configuration that applies to Coder services. 
 services:
-  # services.nodeSelector -- Node selectors that will be applied to all Coder
-  # managed services.
-  # See: https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector
+  # services.annotations -- A KV mapping of annotations. See:
+  # https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/
+  annotations: {}
```
ericpaulsen, line 128: since `annotations` is now under a general `services` value, what is the syntax for specifying a specific service?
I expect a user would need to define the service first, then include the service-specific annotation after the fact, such as:

```yaml
loadbalancer.cloud.google.com/load-balancer-type: "Internal"
```

or 

```yaml
nginx.ingress.kubernetes.io/ssl-redirect: "true"
```

but would like to confirm.

values.yaml
kylecarbs: Yup!

values.yaml
kylecarbs: Since there isn't an ingress anymore, I didn't feel like it was necessary.

Before we used `deploymentAnnotations`, which applied to all Coder services too.

kylecarbs: You did find a bug though... I wasn't respecting the `services.annotations` value in the Coder service. So tyty

kylecarbs: Err I guess I'm confused... how did we do this before to annotate a service? If we didn't that's all good... just curious

ericpaulsen: I'm not exactly sure, I never set annotations outside of the ingress object/service (hence my question above about the syntax). 

Also, it looks like `serviceAccount.annotations` was [used to set annotations for all Coder services](https://github.com/cdr/enterprise-helm/blob/be124a4e5b2839f31834c7ee1fb4b59699b164cc/values.yaml#L383). `deploymentAnnotations` doesn't have any comments about it.

kylecarbs: Going to merge for now... we'll find out much more bugs I'm sure in testing.

PR #108, opened 2021-06-24
chore: update ingress-nginx image

coadler:


jawnsy:
[ch14376]

jawnsy (approved):
LGTM! Please review the upstream changelog to make sure there's nothing we need to worry about/mention to customers: https://github.com/kubernetes/ingress-nginx/blob/controller-v0.47.0/Changelog.md

PR #110, opened 2021-06-29
chore(ci): `exit 1` when `make fmt` is dirty

coadler:


jawnsy (approved):
good catch, thanks for fixing!

jawnsy:
@coadler Can you rebase, remove the `-e` from the Makefile (it doesn't seem necessary on Ubuntu), and merge this? You can also close this and I can take it over if you'd like

https://github.com/cdr/enterprise-helm/blob/043db4cfc140ec880c9afecb144a40deb7408eea/Makefile#L41

PR #114, opened 2021-07-07
fix: Ensure old value is set before applying

kylecarbs:
Fixes a bug where old values are used even if they aren't set.


jawnsy (approved):
Seems reasonable

PR #116, opened 2021-07-07
Fix missing rules on replica ingress

f0ssel:
This is the reason the tls isn't working out of the box

### DIFF
--- a/templates/ingress.yaml
+++ b/templates/ingress.yaml
@@ -363,7 +363,21 @@ spec:
       name: {{ include "coder.serviceName" . }}
       port:
         number: 8080
-  {{- if not .Values.coderd.replica.enable }}
+  {{- if .Values.coderd.replica.enable }}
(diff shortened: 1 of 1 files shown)


templates/ingress.yaml
```diff
@@ -363,7 +363,21 @@ spec:
       name: {{ include "coder.serviceName" . }}
       port:
         number: 8080
-  {{- if not .Values.coderd.replica.enable }}
+  {{- if .Values.coderd.replica.enable }}
```
jawnsy, line 366: I'm not clear what this change means, exactly -- it seems like the case before would've defaulted to coder/8080? How come the explicit rule is needed?

f0ssel: the automatic tls hook with letsencrypt doesn't work unless you specify a rule it seems

deansheather (approved):

templates/ingress.yaml
jawnsy: Ah, TIL! thanks for the info

PR #118, opened 2021-07-07
Replica access URL + delete dashboard deployment

deansheather:


templates/dashboard.yaml
```diff
@@ -1,75 +1,7 @@
...
-            httpGet:
-              path: /healthz
-              port: 3000
-            initialDelaySeconds: 10
-            failureThreshold: 7
-            periodSeconds: 10
-{{- include "coder.resources" .Values.dashboard.resources | indent 10 }}
+# The dashboard service only exists for compatibility with custom ingresses.
+# Dashboard is no longer needed, so we point it's traffic directly at the coderd
+# pods now.
```
kylecarbs, line 4: This will output since it uses `#`. We should use `{{/* */}}` instead.

values.yaml
```diff
@@ -39,7 +39,12 @@ coderd:
     # Replicas enable low-latency access to workspaces all over the world. Read more:
     # TODO: Link to docs.
     enable: false
-    # coderd.replica.primaryURL -- URL of the primary deployment.
+    # coderd.replica.accessURL -- Access URL of the primary deployment. Must be
```
kylecarbs, line 42: "URL of the replica that clients will connect to."

values.yaml
```diff
@@ -39,7 +39,12 @@ coderd:
...
+    # the URL that clients will connect from.
+    # e.g. https://sydney.coder.myorg.com
+    accessURL: ""
+    # coderd.replica.primaryURL -- Access URL of the primary deployment. Must be
```
kylecarbs, line 46: "URL of the primary Coder deployment. Must be accessible from the replica and clients."

kylecarbs (changes requested):

templates/dashboard.yaml
```diff
@@ -79,16 +13,10 @@ metadata:
 spec:
   type: {{ .Values.serviceType | quote }}
   selector:
-    coder.deployment: dashboard
+    coder.deployment: {{ include "coder.serviceName" . }}
```
kylecarbs, line 16: Wouldn't this still be the dashboard?

kylecarbs (approved):

templates/dashboard.yaml
deansheather: This changes which pods the service points to. We want it to point to the coderd pods since that's where the dashboard is

PR #121, opened 2021-07-10
feat: configurable pod and container security contexts

jawnsy:
I tested this in dev-3, seems to work fine. I haven't touched the other containers, since we're phasing them out anyways.

More info about seccomp: https://kubernetes.io/docs/tutorials/clusters/seccomp/

The default option isn't very restrictive, but is better than nothing and suitable for most applications.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,19 +22,26 @@ View [our docs](https://coder.com/docs/setup/installation) for detailed installa
 
 | Key                 | Type | Description | Default                         |
 | ------------------- | ---- | ----------- | ------------------------------- |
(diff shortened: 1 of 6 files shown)


deansheather (approved):

jawnsy:
Blah, OpenShift doesn't like setting seccomp by default, so I guess this needs to be controlled in the values file:

```
Error creating: pods "coderd-58776f967c-jlj49" is forbidden: unable to validate against any security context constraint: [pod.metadata.annotations.seccomp.security.alpha.kubernetes.io/pod: Forbidden: seccomp may not be set pod.metadata.annotations.container.seccomp.security.alpha.kubernetes.io/migrations: Forbidden: seccomp may not be set pod.metadata.annotations.container.seccomp.security.alpha.kubernetes.io/coderd: Forbidden: seccomp may not be set]
```

kylecarbs (approved):

deansheather (approved):

jawnsy:
Tested this with OpenShift and GKE, this change makes the pod/container security contexts for coderd fully configurable, so should work everywhere... In a future release, we may want to make only some fields configurable (i.e. there's not really any reason to allow privilege escalation) but this seems fine for now

PR #126, opened 2021-07-13
chore: Add back certs.*

kylecarbs:
This was removed to simplify the Helm. Customers use it, and we can always deprecate it later.


kylecarbs:
[ch14865]

jawnsy (approved):

PR #127, opened 2021-07-19
Add coderd.trustProxyIP helm value

deansheather:
```
  # coderd.trustProxyIP -- Whether Coder should trust X-Real-IP and/or
  # X-Forwarded-For headers from your reverse proxy. This should only be turned
  # on if you're using a reverse proxy that sets both of these headers. This is
  # always enabled if the Nginx ingress is deployed.
```


jawnsy (approved):
Seems fair enough for now. I think we should make this configurable in the UI and remove it from here (making it configurable via a separate ConfigMap or Secret if needed)

deansheather:
We decided not to make it configurable in the UI because the value may differ per deployment.

PR #129, opened 2021-07-20
feat: add values for oidc refresh opt-in

coadler:


templates/coderd.yaml
```diff
@@ -176,6 +176,12 @@ spec:
             - name: SSL_SERVER_KEY_FILE
               value: {{ join ":" $serverKeys }}
             {{- end }}
+            {{- if .Values.coderd.oidc.enableRefresh }}
+            - name: ENABLE_OIDC_REFRESH
```
deansheather, line 180:
```suggestion
            - name: OIDC_ENABLE_REFRESH
```
so they both start with OIDC

deansheather (approved):

templates/coderd.yaml
coadler: true

PR #132, opened 2021-07-22
feat: allow specifying the super admin password via a secret

coadler:
[ch7427]

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -121,6 +121,21 @@ coderd:
     # the user to the OIDC provider.
     redirectOptions: {}
 
+  # Options for customizing the built-in super admin account for managing a
+  # Coder installation.
+  superAdmin:
+    # Options for configuring the secret used to specify the passwod for the
+    # built-in super admin account.
+    passwordSecret:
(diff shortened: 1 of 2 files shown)


values.yaml
```diff
@@ -121,6 +121,21 @@ coderd:
     # the user to the OIDC provider.
     redirectOptions: {}
 
+  # Options for customizing the built-in super admin account for managing a
+  # Coder installation.
+  superAdmin:
+    # Options for configuring the secret used to specify the passwod for the
```
jawnsy, line 127:
```suggestion
    # Options for configuring the secret used to specify the password for the
```

jawnsy (approved):
Seems alright for now. Longer-term I think we should have a mechanism for loading a list of accounts and credentials from YAML, and loading that as a ConfigMap that is loaded on startup

coadler:
> Seems alright for now. Longer-term I think we should have a mechanism for loading a list of accounts and credentials from YAML, and loading that as a ConfigMap that is loaded on startup

Definitely agree. I think this is a good first step. It'd be nice to offer additional controls for the super-admin account, like being able to disable it.

PR #133, opened 2021-07-24
chore: add minimum Kubernetes version requirement

jawnsy:
Enforce the minimum Kubernetes version in Helm chart, according
to our version support policy.

--

When using an older version of Kubernetes, the error will look like this:

```shell-session
$ helm template --kube-version=1.18.1 coder .
Error: chart requires kubeVersion: >= 1.19.0 which is incompatible with Kubernetes v1.18.1
```

Also verified with `kind`:

```shell-session
$ kubectl get node
NAME                  STATUS   ROLES    AGE   VERSION
coder-control-plane   Ready    master   52s   v1.18.2
$ helm install coder . --namespace=coder
Error: chart requires kubeVersion: >= 1.19.0 which is incompatible with Kubernetes v1.18.2
```

### DIFF
--- a/Chart.yaml
+++ b/Chart.yaml
@@ -6,6 +6,13 @@ name: coder
 description: Coder moves developer workspaces to your cloud and centralizes their creation and management.
 appVersion: master
 version: 1.19.0
+# Coder follows the Kubernetes upstream version support policy, and the
+# latest stable release version of Coder supports the previous two minor
+# releases as well as the current release of Kubernetes at time of
+# publication.
+#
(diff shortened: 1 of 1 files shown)


deansheather (approved):

kylecarbs:
Seems like this doesn't work with GKE...

```
rem ~/P/c/enterprise-helm (service-annotations *• =) helm install --dry-run coder ./
Error: chart requires kubeVersion: >= v1.19 which is incompatible with Kubernetes v1.19.10-gke.1600
```

jawnsy:
@kylecarbs Ahh, that's a bummer.

@deansheather Thanks for the revert

PR #134, opened 2021-07-27
Add service annotations to coderd service

deansheather:


templates/coderd.yaml
```diff
@@ -240,6 +240,10 @@ metadata:
   name: {{ include "coder.serviceName" . }}
   {{- end }}
   namespace: {{ .Release.Namespace | quote }}
+  annotations:
+  {{- range $key, $value := .Values.coderd.serviceAnnotations }}
+    {{ $key }}: {{ $value | quote }}
+  {{- end }}
```
deansheather, line 246: Is this OK if there's nothing in the map?

kylecarbs: Should be

kylecarbs: I'll try this

kylecarbs (approved):
This works, but you should remove the minimum Kube version added in #133. Seems like that isn't compatible with GKE.

PR #138, opened 2021-07-29
Allow static NodePort

Emyrk:
# What this does

We need this to get Kind to work for 1.21. It allows setting static nodeports


README.md
```diff
@@ -1,4 +1,4 @@
-<!-- DO NOT EDIT. THIS IS GENERATED FROM README.md.gotmpl -->
+-e <!-- DO NOT EDIT. THIS IS GENERATED FROM README.md.gotmpl -->
```
kylecarbs, line 1: -e?

values.yaml
```diff
@@ -21,6 +21,18 @@ coderd:
...
     # the cloud provider does not support this feature.
     loadBalancerSourceRanges: []
+
+  # coderd.serviceNodePorts -- Allows manually setting static node ports for the coderd service.
+  # This is only helpful if static ports are required, and usually should be left alone.
+  # By default these are dynamically chosen.
+  serviceNodePorts:
+    # coderd.serviceNodePorts.coderd -- Sets a static 'coderd' service non-TLS nodePort.
+    # This should usually be omitted.
+    coderd: 30080
```
kylecarbs, line 31: Since this is nested under coderd already let's do http and https

kylecarbs (changes requested):

README.md
Emyrk: straight up typo. Whooops

Emyrk: @kylecarbs Apparently the `make fmt` placed that `-e`

jawnsy: I'm not sure if enterprise-dev on arch behaves differently from Ubuntu (enterprise-dev-ubuntu), that might explain the difference here? The -e was part of the initial PR :man_shrugging:

values.yaml
```diff
@@ -21,6 +21,18 @@ coderd:
     # will be restricted to the specified client IPs. This field will be ignored if
     # the cloud provider does not support this feature.
     loadBalancerSourceRanges: []
+
+  # coderd.serviceNodePorts -- Allows manually setting static node ports for the coderd service.
+  # This is only helpful if static ports are required, and usually should be left alone.
+  # By default these are dynamically chosen.
+  serviceNodePorts:
```
jawnsy, line 28: By adding these to the values file, these will be the default values applicable to all installs - is that what we want? What's the default behavior (e.g. if nodePort isn't specified)?

Emyrk: The `values.yaml` is the default?
The original default is that the nodePorts are selected dynamically. I thought `values.yaml` was an example config for generating the docs.

jawnsy: the values file is the defaults that will be used if you don't specify any that override them, it's included in the Helm chart package (so if you do: `helm install . coder` then it'll use the `values.yaml` file unmodified). otherwise, if you specify --set, then it'll merge the default values file with the args you pass

Emyrk: TIL. So we can't document a field without explicitly setting a default?
I wonder if I can set it to `null`?

I prefer the default to be unset.

Emyrk: I could also make the value "0" effectively be "unset"

jawnsy: Yeah I think null should work

jawnsy: You can test by running `make lint/kubernetes`, this will generate a bunch of YAML files somewhere that you can examine

Emyrk: `null` works

kylecarbs (approved):

PR #139, opened 2021-07-29
Separate ingress.usedefault from .enable

mterhar:
This change allows ingress.enabled to create the ingress object for
use with another ingress controller. 
used in conjunction with coderd.serviceSpec.type=ClusterIP,
coderd.devurlsHost, and ingress.host to get the full ingress experience

### DIFF
--- a/templates/ingress.yaml
+++ b/templates/ingress.yaml
@@ -335,6 +335,8 @@ spec:
       protocol: TCP
 {{- end }}
 ---
+{{- end }}
+{{- if (merge .Values dict | dig "ingress" "enable" true) }}
 apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:


templates/coderd.yaml
```diff
@@ -230,7 +230,7 @@ data:
   privateKey: {{ genPrivateKey "ecdsa" | b64enc }}
   {{- end }}
 {{- end }}
-{{ if eq (include "coder.useServiceNext" .) "true" }}
+{{- if and ( eq (include "coder.useServiceNext" .) "true"  ) ) ( eq (merge .Values dict | dig "ingress" "enable" false) ) }}
```
mterhar, line 233: ``
```suggestion
{{- if and ( eq (include "coder.useServiceNext" .) "true"  ) ) (merge .Values dict | dig "ingress" "enable" false) }}
```

templates/coderd.yaml
```diff
@@ -230,7 +230,7 @@ data:
   privateKey: {{ genPrivateKey "ecdsa" | b64enc }}
   {{- end }}
 {{- end }}
-{{ if eq (include "coder.useServiceNext" .) "true" }}
+{{- if and ( eq (include "coder.useServiceNext" .) "true"  ) ) (merge .Values dict | dig "ingress" "enable" false) }}
```
mterhar, line 233: ``
```suggestion
{{- if and ( eq (include "coder.useServiceNext" .) "true"  ) (merge .Values dict | dig "ingress" "enable" false) }}
```

mterhar:
I'm not convinced that this isn't causing more trouble elsewhere.  The useServiceNext mechanism seem to tie into making envproxy a thing and then the ingress gets all complicated again. 

I'm making these changes to support the istio deployment approach for Platform One so their helm chart is going to have some differences (as it always does) but I had hoped to bring that flexibility back to our chart.  Once I get the platform one chart merged, i'll revisit this.

templates/coderd.yaml
```diff
@@ -230,7 +230,7 @@ data:
   privateKey: {{ genPrivateKey "ecdsa" | b64enc }}
   {{- end }}
 {{- end }}
-{{ if eq (include "coder.useServiceNext" .) "true" }}
+{{- if and ( eq (include "coder.useServiceNext" .) "true"  ) (merge .Values dict | dig "ingress" "enable" false) }}
```
mterhar, line 233:
```suggestion
{{- if and ( eq (include "coder.useServiceNext" .) "true"  ) ( eq (merge .Values dict | dig "ingress" "enable" false) false ) }}
```

This actually resolves the ingress.enable to the proper value to skip the fancy LB.

mterhar:
fresh attempt in #144

PR #142, opened 2021-08-03
remove envproxy references from chart

f0ssel:


kylecarbs (approved):
LGTM. We can remove the entire Ingress object now too!

deansheather:
We can't remove the ingress object because it'll break customers that use cert-manager with ingress annotations, which the docs have recommended in the recent past

deansheather (approved):

jawnsy:
We also need an ingress because LoadBalancer controllers are not a mandatory part of Kubernetes, so not every cluster has them. The clouds all have load balancer controllers but bare metal systems might not, which is why the [MetalLB documentation](https://metallb.universe.tf/) says:

>Kubernetes does not offer an implementation of network load balancers (Services of type LoadBalancer) for bare-metal clusters. The implementations of network load balancers that Kubernetes does ship with are all glue code that calls out to various IaaS platforms (GCP, AWS, Azure…). If you’re not running on a supported IaaS platform (GCP, AWS, Azure…), LoadBalancers will remain in the “pending” state indefinitely when created.

f0ssel:
Yeah I think keeping an optional, dumb one is a happy medium

PR #143, opened 2021-08-09
chore: set app label in coderd deployment

coadler:
Cilium seems to check this label when identifying services
image

### DIFF
--- a/templates/coderd.yaml
+++ b/templates/coderd.yaml
@@ -3,6 +3,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   labels:
+    app: {{ include "coder.serviceName" . }}
     coder.deployment: {{ include "coder.serviceName" . }}
   name: {{ include "coder.serviceName" . }}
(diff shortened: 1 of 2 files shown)


templates/coderd.yaml
```diff
@@ -3,6 +3,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   labels:
+    app: {{ include "coder.serviceName" . }}
```
jawnsy, line 6: ah, I wonder if it's related to https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

jawnsy (approved):
LGTM

templates/coderd.yaml
coadler: hmm maybe we should set some of these as well

jawnsy: I think Helm adds a bunch. Not a big deal, can always do it later!

jawnsy (approved):

PR #144, opened 2021-08-12
Load balancer TLS termination

mterhar:
If TLS is terminated by an ALB, the service port 80 routes traffic without redirecting
to HTTPS and the HTTPS port presents an error since it's pointing to another HTTPS
port of the node.

This change allows the load balancer to expose only port 443 externally which routes to
the HTTP port of the coderd pod.

### DIFF
--- a/values.yaml
+++ b/values.yaml
@@ -36,6 +36,11 @@ coderd:
   # coderd.serviceAnnotations -- Extra annotations to apply to the coderd service.
   serviceAnnotations: {}
 
+  # coderd.httpsToHttp -- eliminates the external http port and routes traffic from
+  # the external https port to the internal http port. Useful for when the load balancer
+  # performs TLS termination (like Amazon's ACM)
+  httpsToHttp: false
+
   # coderd.trustProxyIP -- Whether Coder should trust X-Real-IP and/or
   # X-Forwarded-For headers from your reverse proxy. This should only be turned
   # on if you're using a reverse proxy that sets both of these headers. This is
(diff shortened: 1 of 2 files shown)


Emyrk (approved):
I don't have much context on this, but it seems reasonable.

values.yaml
```diff
@@ -36,6 +36,11 @@ coderd:
   # coderd.serviceAnnotations -- Extra annotations to apply to the coderd service.
   serviceAnnotations: {}
 
+  # coderd.httpsToHttp -- eliminates the external http port and routes traffic from
+  # the external https port to the internal http port. Useful for when the load balancer
+  # performs TLS termination (like Amazon's ACM)
+  httpsToHttp: false
```
jawnsy, line 42: I've done a little bit of reading ([How do I terminate HTTPS traffic on Amazon EKS workloads with ACM?
](https://aws.amazon.com/premiumsupport/knowledge-center/terminate-https-traffic-eks-acm/)) but am by no means an expert on AWS.

It seems unusual to me that we'd be using a `LoadBalancer` spec when a frontend device is performing termination for us, and instead I'd have expected that we'd use an `Ingress` definition to do this? A Kubernetes LoadBalancer would be more like an Layer 3 (Network Load Balancer) and a Service with Ingress would be more like a Layer 7 device (Application Load Balancer)

Alternatively, is there a way to configure things manually so that we expose coderd as a ServiceIP (on the internal VPC network) and then manage the NLB/ALB outside of our Kube specs? Adding stuff to our Helm chart, especially in the absence of tests, increases complexity, so my preference is always to keep one-off customizations like this out of there, unless this is a very general thing with all of the clouds...

jawnsy: It seems like an alternative to this implementation would be to make the full service spec configurable as users want, instead of using conditionals based on this `httpsToHttp` setting

PR #146, opened 2021-08-24
add example for custom ingress

mterhar:
I'm using the following command to ensure the right objects are being created: 

`helm template coder . -f examples/ingress/ingress.values.yaml --set postgres.default.enable=false`

The goal of this is to:

- [x] Suggest values for when deploying into a cluster that can't make more LBs
- [x] Add a simple ingress object that will exist when `ingress.enable=true`
- [x] Support TLS-at-the-ingress with annotations and such

### DIFF
--- a/examples/ingress/ingress.values.yaml
+++ b/examples/ingress/ingress.values.yaml
@@ -4,6 +4,7 @@
 # by creating a LoadBalancer service
 coderd:
   serviceNext: true
+  devurlsHost: '*.devurls.coderhost.com'
   serviceSpec:
     type: ClusterIP
     # The values.yaml file in the chart includes LoadBalancer
@@ -12,6 +13,12 @@ coderd:
     loadBalancerIP: null
     externalTrafficPolicy: null
     loadBalancerSourceRanges: null
-# Add your own ingress spec based on the type of ingress you
-# have deployed in your cluster. Send all traffic to the 
-# coderd service on ports 80, and 443.
+# Add the ingress values section to enable the ingress resource
+# without the controller
+ingress:
+  # Enable set to true creates the ingress resource
+  enable: true
+  # Ingress needs a host name so it can share a controller
+  host: 'coder.coderhost.com'
+  # useDefault set to false disables creation of the ingress controller
+  useDefault: false
--- a/examples/ingress/ingress.values.yaml
+++ b/examples/ingress/ingress.values.yaml
@@ -1,0 +1,17 @@
+# Using coder with ingress in versions from 1.21 and newer
+# Coder's built-in ingress controller is no longer packaged
+# the coderd pod does not require a fanout so it is exposed
+# by creating a LoadBalancer service
+coderd:
+  serviceNext: true
+  serviceSpec:
+    type: ClusterIP
+    # The values.yaml file in the chart includes LoadBalancer
+    # specs which need to have the keys removed using null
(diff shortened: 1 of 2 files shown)


mterhar:
@sreya I have to add a bunch of this stuff to platform one's big bang deployment since it only has 1 LB which is shared by all of their services. The ingress stuff is required for them even with serviceNext.

templates/ingress.yaml
```diff
@@ -409,3 +409,38 @@ spec:
...
+  name: coderd-ingress
+  namespace: {{ .Release.Namespace | quote }}
+  annotations:
+    nginx.ingress.kubernetes.io/proxy-body-size: "0"
+  {{- range $key, $value := merge .Values dict | dig "ingress" "annotations" dict }}
+    {{ $key }}: {{ $value | quote }}
+  {{- end }}
+spec:
+{{- include "coder.ingress.tls" . }}
+  rules:
```
jawnsy, line 425: suggest we add a defaultBackend too: https://kubernetes.io/docs/concepts/services-networking/ingress/#default-backend

templates/ingress.yaml
```diff
@@ -409,3 +409,38 @@ spec:
...
 {{- end }}
 ---
+{{- if and (merge .Values dict | dig "coderd" "serviceNext" true) (merge .Values dict | dig "ingress" "enable" true) }}
+apiVersion: networking.k8s.io/v1
+kind: Ingress
+metadata:
+  name: coderd-ingress
+  namespace: {{ .Release.Namespace | quote }}
+  annotations:
+    nginx.ingress.kubernetes.io/proxy-body-size: "0"
```
jawnsy, line 419: I think this can be put in the default values file, then we can make all of the annotations configurable, similar to how we do it with securityContext: https://github.com/cdr/enterprise-helm/blob/e158a751b9a1c3198adff2c08b4a5f50c1e33fdf/templates/coderd.yaml#L36-L42

right now there's no way to clear the `nginx.ingress.kubernetes.io/proxy-body-size` annotation here

FWIW, I think this resolves: https://app.clubhouse.io/coder/story/14644/route-traffic-to-coderd-using-ingress -- thank you!

jawnsy (approved):
LGTM. I'll be more careful to check the preview this time before merging :sweat_smile:

examples/ingress/ingress.values.yaml
```diff
@@ -0,0 +1,26 @@
...
+# without the controller
+ingress:
+  # Enable set to true creates the ingress resource
+  enable: true
+  # Ingress needs a host name so it can share a controller
+  host: 'coder.coderhost.com'
+  # useDefault set to false disables creation of the ingress controller
+  useDefault: false
+  # add annotations for TLS issuers and such
+  annotations: {}
```
jawnsy, line 26: not mandatory but would be nice to modify the OpenShift one to add an example that creates an OpenShift Route (it needs some different annotations)

https://docs.okd.io/latest/networking/routes/route-configuration.html#nw-ingress-creating-a-route-via-an-ingress_route-configuration

templates/ingress.yaml
mterhar: Default backend is supposed to be on the controller rather than the ingress resources (by convention)... this choice was deliberate.

jawnsy:
[ch14644]

PR #152, opened 2021-09-16
Upgrade workspace provider helm chart so it can be uninstalled

f0ssel:
So after 1.22 you no longer need the envproxy deployment, but if you uninstall the helm chart it removes the RBAC resources and kills the workspace provider. With this change, customers can upgrade to 1.22+ they can safely uninstall the chart afterwards. 

@deansheather I'll need some advising on how this helm chart can get deployed in 1.23 and if we can backport it to 1.22


deansheather (approved):
I can manually build and deploy this to the helm repo

f0ssel:
sick

f0ssel:
@sreya looping you in, this is the change we discussed for big P

PR #156, opened 2021-09-27
Fix: port mismatch ingress to service

mterhar:
Ingress port needs to match the port number on the coderd service.

@deansheather This one.

### DIFF
--- a/templates/ingress.yaml
+++ b/templates/ingress.yaml
@@ -432,7 +432,7 @@ spec:
           service:
             name: {{ include "coder.serviceName" . }}
             port:
-              number: 80
+              number: 8080
   - host: {{ merge .Values dict | dig "coderd" "devurlsHost" "" | quote }}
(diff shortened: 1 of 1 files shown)


mterhar:
Swtiched to using the port name which is consistent whether the port number is 80 or 8080. 

Validation steps: 

```bash
helm pull coder/coder --version 1.23.0
tar zvfx coder-1.23.0.tgz
mv coder coder-1.23
```

Replace the contents of templates/ingress.yaml

```bash
kubectl create ns coder-ing-test
helm upgrade -i coder coder-1.23 -n coder-ing-test -f coder-1.23/examples/ingress/ingress.values.yaml
```

I used coder.ing.terharlab.com to validate the connectivity and it worked just fine with the port name on K8s 1.20.9 in GCP.

deansheather (approved):

PR #158, opened 2021-11-02
feat: add unit tests for Helm chart

jawnsy:
Very rough first pass, mostly a proof-of-concept. The tests aren't complete, and not part of CI yet, but the goal is to show what writing them could look like.


tests/values.go
```diff
@@ -0,0 +1,245 @@
...
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/apimachinery/pkg/util/yaml"
+)
+
+var _ = fmt.Stringer(CoderValues{})
+
+// CoderValues is a typed Go representation of Coder's values file,
+// suitable for writing tests. This provides code completion for Go
+// tests.
+type CoderValues struct {
```
johnstcn, line 30: Keeping this in sync manually is going to be an issue down the line.
Could we generate this automatically from `values.yaml`? 

Aside: _is_ there any decent tool for doing this directly? I found [yaml2go](https://github.com/PrasadG193/yaml2go) but it appears to not be maintained. We could also go the route of YAML -> JSON -> Go, for which there is a [slightly more up to date tool](https://github.com/m-zajac/json2go). If all else fails, I suppose there's Swagger.

johnstcn (approved):
I really like the shape of this. Being able to just take a `values.yaml`, render the Helm chart, and make assertions about the output is really powerful!

tests/values.go
jawnsy: I discussed this with @Emyrk, there's a way you can define a JSON Schema to validate the values, and we could generate both the values.yaml and these structs from that. But in the interest of an initial implementation, I left taht out for now :)

johnstcn: Nice :-) I think this will really help!

PR #160, opened 2021-11-05
chore: update dependencies

jawnsy:
* Update kube-linter from 0.2.2 to 0.2.5
* Add gotestsum to summarize test output
* Add golangci-lint to lint test code

### DIFF
--- a/scripts/install_deps.sh
+++ b/scripts/install_deps.sh
@@ -6,7 +6,7 @@ set -euo pipefail
 cd "$(dirname "$0")"
 source "./lib.sh"
 
-TMPDIR="$(mktemp -d)"
+TMPDIR=$(mktemp -d)
 BINDIR="/usr/local/bin"
 
 curl_flags=(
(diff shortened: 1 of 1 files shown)


scripts/install_deps.sh
```diff
@@ -18,22 +18,40 @@ curl_flags=(
 run_trace false sudo apt-get install --no-install-recommends --yes \
   shellcheck
 
-run_trace false curl "${curl_flags[@]}" "https://get.helm.sh/helm-v3.7.1-linux-amd64.tar.gz" \| \
-  tar -C "$TMPDIR" --strip-components=1 -zxf - linux-amd64/helm
+# gotestsum makes test output more readable
+GOTESTSUM_VERSION=1.7.0
```
johnstcn, line 22: nit: (non-blocking): I find that puting `XXX_VERSION` at the very top of the file, all clustered together, makes it somewhat easier. My personal preference though entirely.

johnstcn (approved):
👍

coadler (approved):

scripts/install_deps.sh
jawnsy: I can see that, yeah - I guess the downside is that separating them makes it easier to remove code without removing corresponding env vars :thinking: I think I'll leave it as-is for now, but I also don't feel strongly about it and might change my mind later lol

I actually want to move all of these to an image build (and build our stuff inside an image), so this script is kinda throwaway code

PR #168, opened 2021-11-09
Add service account resource to rbac

f0ssel:


deansheather (approved):

templates/rbac.yaml
```diff
@@ -30,7 +30,7 @@ metadata:
   namespace: {{ .Release.Namespace | quote }}
 rules:
   - apiGroups: ["", "apps", "networking.k8s.io"] # "" indicates the core API group
-    resources: ["persistentvolumeclaims", "pods", "deployments", "services", "secrets", "pods/exec","pods/log", "events", "networkpolicies"]
+    resources: ["persistentvolumeclaims", "pods", "deployments", "services", "secrets", "pods/exec","pods/log", "events", "networkpolicies", "serviceaccounts"]
```
jawnsy, line 33: Probably need to document this requirement, yeah? And need to modify the doctor too

f0ssel: Good call on doctor, will circle back on that and yeah need I will be making a notice in the release notes on the permission change. Even if missing it will still work OK, the feature that uses these permissions are completely opt-in.

jawnsy: Alternatively, we can make it a new value, default it to false, and only set this if that value is true -- this is only needed if someone wants to use Workload Identity or the similar EKS feature, right?

The fewer RBAC permissions we need, the less trouble we'll have with installations. The user doing the install has to have all the permissions that they're granting to the coderd service account, if I understand correctly

jawnsy: @johnstcn might be able to help with changes to the doctor and @khorne3 can help with listing stuff in release notes

kylecarbs: Should this be an option in the Helm? Since it's not applicable to all deployments, it seems like an arbitrary addition.

f0ssel: That's correct, it's not required unless you want to use the workspace service account annotation feature, so I'm fine putting this behind an option if we'd like.

kylecarbs: I think that'd be wise.

PR #169, opened 2021-11-13
chore: remove compatibility for legacy networking

jawnsy:
I've tagged a bunch of people for review because this is a pretty big change. I have not tested this yet, but it does at least pass the kind integration and very limited unit tests we have... Recommendations for a test strategy would be most welcome, I was just going to give it a smoke test on dev-4.


templates/_common.tpl
```diff
@@ -94,34 +93,25 @@ volumeMounts:
   coder-managed services.
 */}}
 {{- define "coder.serviceTolerations" }}
-{{- if ne (include "movedValue" (dict "Values" .Values "Key" "services.tolerations")) "" }}
-tolerations:
-{{ include "movedValue" (dict "Values" .Values "Key" "services.tolerations") }}
+{{- if .Values.services.tolerations }}
+tolerations: {{ .Values.services.tolerations }}
```
deansheather, line 97: does this work?

templates/_common.tpl
```diff
@@ -94,34 +93,25 @@ volumeMounts:
...
 {{- define "coder.accessURL" }}
-{{- if .Values.cemanager }}
-{{- if ne (merge .Values dict | dig "cemanager" "accessURL" "") "" }}
-{{- .Values.cemanager.accessURL -}}
-{{- else -}}
-    http://cemanager.{{ .Release.Namespace }}{{ include "movedValue" (dict "Values" .Values "Key" "services.clusterDomainSuffix") }}:8080
-{{- end }}
-{{- else -}}
-{{- if ne (merge .Values dict | dig "coderd" "accessURL" "") "" }}
+{{- if .Values.coderd.accessURL "" }}
```
deansheather, line 104:
```suggestion
{{- if ne .Values.coderd.accessURL "" }}
```

templates/coderd.yaml
```diff
@@ -9,12 +9,9 @@ metadata:
     coder.deployment: {{ include "coder.serviceName" . }}
   name: {{ include "coder.serviceName" . }}
   namespace: {{ .Release.Namespace | quote }}
-  annotations:
-  {{- range $key, $value := fromYaml (include "movedValue" (dict "Values" .Values "Key" "services.annotations")) }}
-    {{ $key }}: {{ $value | quote }}
-  {{- end }}
+  annotations: {{- toYaml .Values.services.annotations | nindent 4 }}
```
deansheather, line 12: I think this needs to be on it's own line and without the `-` prefix

templates/coderd.yaml
```diff
@@ -132,20 +111,15 @@ spec:
...
             - name: DEVURL_HOST
-              value: {{ include "movedValue" (dict "Values" .Values "Key" "coderd.devurlsHost") }}
+              value: {{ .Values.coderd.devurlsHost | quote }}
             - name: POD_IP
               valueFrom:
                 fieldRef:
                   fieldPath: status.podIP
             - name: VERBOSE
               value: "true"
-            {{- if and (eq (include "coder.hasNginxIngress" .) "true") (eq (merge .Values dict | dig "coderd" "serviceNext" false) false) (merge .Values dict | dig "ingress" "enable" true) (merge .Values dict | dig "ingress" "useDefault" true) }}
```
deansheather, line 142: Shouldn't this still be on if `ingress.enabled`?

templates/coderd.yaml
```diff
@@ -232,21 +206,14 @@ data:
...
-{{ if eq (include "coder.useServiceNext" .) "true" }}
 ---
 apiVersion: v1
 kind: Service
 metadata:
-  {{- if (lookup "v1" "Service" .Release.Namespace "ingress-nginx") }}
-  name: ingress-nginx
-  {{- else }}
   name: {{ include "coder.serviceName" . }}
-  {{- end }}
```
deansheather, line 244: I don't think we can ever get rid of this without causing load balancer IPs to change for all customers

deansheather (commented):
This looks ok but the removal of `nginx-ingress` load balancer is worrying. Also the inconsistent indenting/newlines around `toYaml` should be fixed

templates/_common.tpl
jawnsy: Ah, it would not. Good catch!

templates/coderd.yaml
jawnsy: the `nindent` is supposed to add a newline first, I think. what does the `-` prefix do?

templates/_common.tpl
jawnsy: I guess this could be `{{- if .Values.coderd.accessURL }}` since the empty value is false? not really sure (and that value is missing from the default values.yaml :thinking:)

templates/coderd.yaml
deansheather: Actually in that case this is probably fine as long as you're consistent with this style (`-` prefix and `nindent`)

The `-` prefix or suffix after the `{{` or before the `}}` means "eat all whitespace before or after this block".

templates/_common.tpl
deansheather: yeah that should work

templates/coderd.yaml
jawnsy: Will add it back. I think we should recommend that people add a reserved static IP and add the appropriate annotation though, to be safe

templates/coderd.yaml
mterhar: I don't think we care if they're using our included ingress or not. Letting them trust their own headers will be good for some customers since they have other teams running WAFs/proxies/etc.

mterhar (approved):
Looks great. I think this deserves some sort of new version or name or something when we release it.

kylecarbs:
Why is the removal of nginx-ingress worrying @deansheather?

deansheather:
@kylecarbs the old load balancer service is called ingress-nginx, so we can't change the name of it for existing deployments or load balancer IPs will change

kylecarbs:
Oh I see. Yea, that is sketchy.

templates/coderd.yaml
```diff
@@ -257,7 +232,7 @@ spec:
       targetPort: 8080
       protocol: TCP
       {{ if .Values.coderd.serviceNodePorts.http }}
-      nodePort: {{  .Values.coderd.serviceNodePorts.http  }}
+      nodePort: {{  .Values.coderd.serviceNodePorts.http }}
```
coadler, line 235:
```suggestion
      nodePort: {{ .Values.coderd.serviceNodePorts.http }}
```

coadler (approved):
I like the use of `nindent`. Makes it much more readable

templates/coderd.yaml
jawnsy: nice, good to know, thanks!

mterhar:
> kylecarbs the old load balancer service is called ingress-nginx, so we can't change the name of it for existing deployments or load balancer IPs will change

Can we remove the ingress-nginx loadbalancer from control of helm kinda like we did for the workspace provider service account?  I think we can abandon that object since the future paradigm is that they manage their own LB.

templates/coderd.yaml
jawnsy: if the ingress is enabled, I guess we force trust proxy IP to on - is that the desired behavior?

jawnsy: seems probably safer to just require the user to configure trustProxyIP themselves, even if they've enabled the ingress, so maybe that's the move here?

jawnsy:
I tested this manually in our dev cluster, and since the release has been cut, I think this is safe to merge, so that we get more testing done in preparation for the next release.

PR #166, opened 2021-11-08
feat: add helm test for ingress example

jsjoeio:
This PR adds a new helm test for the [ingress example](https://github.com/cdr/enterprise-helm/blob/main/examples/ingress/ingress.values.yaml).

Fixes #165

PR #174, opened 2021-11-22
chore: update example tests

jawnsy:
Update tests for example values files to coalesce values using
the built-in Helm handling, rather than our custom rendering.
We should only use the struct-based rendering for unit tests.


deansheather (approved):

jawnsy:
@jsjoeio This should unblock #166 once merged

jsjoeio (approved):
Hooray! This looks great 🚀

PR #176, opened 2021-12-05
cleanup: Execute coderd directly

kylecarbs:
The entrypoint script previously set environment variables as
flags to coderd. coderd now consumes the environment variables
directly, removing the need for the script.

There were a few variables that didn't match up, and I'll be creating
a PR to fix those.


jawnsy (approved):
Nice

PR #177, opened 2021-12-06
feat: promote ingress settings to public

jawnsy:
* Add ingress section to values.yaml and re-generate documentation
* Add tests for ingress settings
* Change comment syntax to {{/* */}} to prevent it from appearing in the generated Ingress spec


jawnsy:
This has some of the changes in #169 (namely, exposing the ingress in our docs), but I split it into a separate PR to make it less risky and easier to review

tests/ingress_test.go
```diff
@@ -0,0 +1,170 @@
...
+			AssertFunc: func(t *testing.T, ingress *netv1.Ingress) {
+				defaultAnnotations := map[string]string{
+					"nginx.ingress.kubernetes.io/proxy-body-size": "0",
+				}
+				require.Equal(t, defaultAnnotations, ingress.Annotations)
+				pathTypePrefix := netv1.PathTypePrefix
+				expectedRules := []netv1.IngressRule{
+					{
+						Host: "install.coder.com",
+						IngressRuleValue: netv1.IngressRuleValue{
```
johnstcn, line 37: readability (non-blocking): Could you extract this struct to a variable?

tests/ingress_test.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/jinzhu/copier"
+	"github.com/stretchr/testify/require"
+)
+
+func TestIngress(t *testing.T) {
+	chart := LoadChart(t)
+
+	tests := []struct {
+		Name       string
+		ValuesFunc func(v *CoderValues)
```
johnstcn, line 18: nit (non-blocking): Could you add a comment describing what this does?

johnstcn (approved):
I haven't tested this, but it looks fine to me. Some minor comments.

Emyrk (commented):
Those tests look tedious to write!

tests/ingress_test.go
jawnsy: I'm confused, this whole struct is already a variable. Can you show me an example of what you mean?

johnstcn: This is basically what I was thinking; all the values look to be the same:
```
var tcpCoderdRoot = netv1.IngressRuleValue{
    ...
}

...

expectedRules := []netv1.IngressRule{
    {
        Host: "xxx",
        IngressRuleValue: tcpCoderdRoot,
        ...
    },
...
```

jawnsy: Extracted out some of the shared stuff, thanks!

tests/ingress_test.go
jawnsy: added!

jawnsy:
>Those tests look tedious to write!
@Emyrk They are kinda tedious, yeah :frowning: But less tedious than always testing things manually, I think!

Do you have ideas for how to improve/simplify the API or test code? I'm open to suggestions!

I'm not in love with this code, but it's better than the alternative of manual tests only

jawnsy:
Tested this in dev-4 and everything deploys fine, should be safe to merge

PR #178, opened 2021-12-08
feat: support postgres mtls

coadler:


values.yaml
```diff
@@ -201,6 +201,27 @@ postgres:
   # the PostgreSQL connection. For acceptable values, see:
   #  https://www.postgresql.org/docs/9.1/libpq-ssl.html
   sslMode: "require"
+  # postgres.ssl -- Options for configuring the SSL cert, key, and root cert
+  # when connecting to Postgres.
+  ssl:
```
jawnsy, line 206: what do you think about using `tls` here, makes things consistent with `ingress.tls` etc

templates/_common.tpl
```diff
@@ -64,15 +76,30 @@ volumes:
...
 
 {{/* 
   coder.volumeMounts adds a volume mounts stanza if a cert.secret is provided.
 */}}
 {{- define "coder.volumeMounts" }}
-{{- if or (merge .Values dict | dig "certs" "secret" "name" false) (ne (include "movedValue" (dict "Values" .Values "Key" "coderd.tls.hostSecretName")) "") }}
 volumeMounts:
-{{- end }}
+  - name: tmp-pgcerts
+    mountPath: /tmp/pgcerts
```
jawnsy, line 102: maybe just mount this as /tmp so that other things can use it?

deansheather (approved):

templates/_common.tpl
deansheather: We can't mount at `/tmp` because that's where the coderd binary is. There's no reason that binary can't be moved in the future, but that's for another PR.

templates/coderd.yaml
```diff
@@ -90,6 +90,7 @@ spec:
             {{- else }}
             {{- toYaml .Values.coderd.securityContext | nindent 12 }}
             {{- end }}
+{{- include "coder.volumeMounts" . | indent 10 }}
```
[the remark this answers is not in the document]
coadler: It didn't need volume mounts before.

values.yaml
coadler: Postgres always refers to it as `SSL` so I was staying consistent with their documentation and naming.

templates/_common.tpl
jawnsy: ah ok

templates/coderd.yaml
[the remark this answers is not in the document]
jawnsy: huh, I would've thought we'd need volume mounts for TLS certs and stuff :man_shrugging:

tests/examples_test.go
```diff
@@ -113,40 +86,58 @@ func TestExamples(t *testing.T) {
...
-						"expected one container")
-					require.Equal(t, test.ContainerSecurityContext,
-						deployment.Spec.Template.Spec.Containers[0].SecurityContext,
-						"expected matching container securityContext")
-
-					break
-				}
-			}
-			require.True(t, found, "expected coderd deployment in manifests")
+			coderd := FindDeployment(t, objs, "coderd")
```
jawnsy, line 130: nice :+1:

tests/values.go
```diff
@@ -60,14 +60,14 @@ type Chart struct {
...
 type CoderValues struct {
-	Certs    *CertsValues    `json:"certs" yaml:"certs"`
-	Coderd   *CoderdValues   `json:"coderd" yaml:"coderd"`
-	Envbox   *EnvboxValues   `json:"envbox" yaml:"envbox"`
-	Ingress  *IngressValues  `json:"ingress" yaml:"ingress"`
-	Logging  *LoggingValues  `json:"logging" yaml:"logging"`
-	Metrics  *MetricsValues  `json:"metrics" yaml:"metrics"`
-	Postgres *PostgresValues `json:"postgres" yaml:"postgres"`
-	Services *ServicesValues `json:"services" yaml:"services"`
+	Certs    *CertsValues    `json:"certs,omitempty" yaml:"certs,omitempty"`
```
jawnsy, line 63: omitempty doesn't really matter here, since these will always be set, based on the defaults in the Chart values.yaml, right?

tests/values.go
```diff
@@ -385,6 +392,15 @@ func (c *Chart) Render(values *CoderValues, options *chartutil.ReleaseOptions, c
 	return objs, nil
 }
 
+func RenderChart(t testing.TB, values *CoderValues) []runtime.Object {
```
jawnsy, line 395: could also make this a func on the Chart object like MustRender or something, I guess

tests/values.go
coadler: When I tested it, it seemed like since it was rendering them all as null, it was overriding the defaults in the Chart `values.yaml`.

tests/values.go
coadler: good idea

tests/values.go
```diff
@@ -498,3 +514,89 @@ func ReadValuesAsMap(path string) (map[string]interface{}, error) {
...
+	decoder := yaml.NewYAMLToJSONDecoder(strings.NewReader(valuesStr))
+	err := decoder.Decode(&values)
+	if err != nil {
+		return nil, fmt.Errorf("error decoding yaml: %w", err)
+	}
+
+	return values, nil
+}
+
+func ReadChartValues(t testing.TB, chart *chart.Chart, valuesStr string) chartutil.Values {
```
jawnsy, line 530: is ReadChartValues and ReadValuesAsMap actually used anywhere? I can't see where they're being used

tests/values.go
```diff
@@ -498,3 +514,89 @@ func ReadValuesAsMap(path string) (map[string]interface{}, error) {
...
+
+	values, err := chartutil.ToRenderValues(chart, valuesMap, DefaultReleaseOptions(), chartutil.DefaultCapabilities.Copy())
+	require.NoError(t, err, "to render values")
+
+	return values
+}
+
+// FindDeployment finds a deployment in the given slice of objects with the
+// given name.
+func FindDeployment(t testing.TB, objs []runtime.Object, name string) *appsv1.Deployment {
```
jawnsy, line 542: maybe want to return a `(*appsv1.Deployment, bool)` here because there's some test cases where we want to validate that the object doesn't exist (e.g. with `ingress.enable = false`, we want to verify that there is no `Ingress` object)

we can always make that adjustment later though

values.yaml
```diff
@@ -218,6 +218,27 @@ postgres:
   # the PostgreSQL connection. For acceptable values, see:
   #  https://www.postgresql.org/docs/9.1/libpq-ssl.html
```
jawnsy, line 219: maybe want to link to the version of the min version of pg we support: https://www.postgresql.org/docs/11/libpq-ssl.html

values.yaml
jawnsy: hmm. alright I guess. I get why they need to call it SSL (backward compatibility) but since this is a new field for us, I prefer TLS. not a big deal to me though.

values.yaml
```diff
@@ -218,6 +218,27 @@ postgres:
...
+  ssl:
+    # postgres.ssl.certSecret -- Secret containing a PEM encoded cert file.
+    certSecret:
+      # postgres.ssl.certSecret.name -- Name of the secret.
+      name: ""
+      # postgres.ssl.certSecret.key -- Key pointing to a certificate in the secret.
+      key: ""
+    # postgres.ssl.keySecret -- Secret containing a PEM encoded key file.
+    keySecret:
+      # postgres.ssl.keytSecret.name -- Name of the secret.
```
jawnsy, line 232: longer-term, we might want to consider an option to load this from a file in the certs directory (e.g. load the file into a Secret and use that secret name automatically)... but this is good for now

jawnsy (approved):
:+1:

tests/values.go
coadler: I used these in a prior iteration when I was embedding a yaml string instead of using the `CoderValues` type. I'll remove these

tests/values.go
coadler: I'll prefix this with a `Must` to leave room for the logic you're describing.

tests/values.go
jawnsy: So a bit of context on the whole CoderValues struct and why it exists... I wanted code completion, because otherwise the Helm API would result in a nested literal situation like:

```go
values := map[string]interface{}{
    "coderd": map[string]interface{}{
        "securityContext": map[string]interface{}{
            "runAsNonRoot": true,
            "seccompProfile": map[string]interface{}{
                "type": "RuntimeDefault",
            },
        },
    },
}
```

This seems kinda ugly and we also lose code completion, as well as validation. I then learned that the reason Helm does it this way (I guess) is that this allows people to distinguish between values being nil (key exists and set to null) and undefined (key not defined).

So the CoderValues always returns the full values defined in the original chart, then the caller needs to copy and modify it, and pass it back. Otherwise, if you just pass a CoderValues struct, then there's no way to distinguish between the unset/undefined (in which case, we want to apply the default) or null (in which case, we want to override the default) cases.

So these things should never be null, otherwise they get passed into Helm as null. If you use omitempty, then there's no way to explicitly set things to null to override defaults either. Either way, we're guessing between one of the cases (unset/undefined, override to null) and in some of the cases our guess will be incorrect (and unfixable) :frowning:

coadler: Would it be better if I just embedded a yaml string then?

tests/pgssl_test.go
```diff
@@ -0,0 +1,70 @@
...
+		}
+
+		objs   = LoadChart(t).MustRender(t, cv)
+		coderd = MustFindDeployment(t, objs, "coderd")
+	)
+
+	for _, vol := range []string{"pgcert", "pgkey", "pgrootcert"} {
+		AssertVolume(t, coderd.Spec.Template.Spec.Volumes, vol, func(t testing.TB, v corev1.Volume) {
+			require.NotNil(t, v.Secret)
+			assert.Equal(t, "pg-certs", v.Secret.SecretName)
```
jawnsy, line 52: just curious: why assert instead of require here?

also, without a message that indicates the current `vol`, it's harder to debug this - consider using Equalf and adding a message with the vol, like: `assert.Equalf(t, "pg-certs", v.Secret.SecretName, "certs do not match for volume %s", vol)`

PR #180, opened 2021-12-10
feat: set soft anti-affinity rule by default

johnstcn:


values.yaml
```diff
@@ -152,26 +152,21 @@ coderd:
...
-  #             matchExpressions:
-  #               - key: app
-  #                 operator: In
-  #                 values:
-  #                   - "coderd"
-  #           topologyKey: kubernetes.io/hostname
-  #         weight: 1
-  # ```
-
+  # deployment. The default rule attempts to schedule multiple coderd
```
jawnsy, line 155: suggest something like:

>Allows specifying an affinity rule for the `coderd` deployment. The default rule prefers to schedule coderd pods on different nodes, which is only applicable if `coderd.replicas` is greater than 1.

values.yaml
```diff
@@ -152,26 +152,21 @@ coderd:
...
+  # pods on different nodes. This only really applies if you set
+  # `coderd.replicas` to a value higher than 1.
+  affinity:
+    podAntiAffinity:
+      preferredDuringSchedulingIgnoredDuringExecution:
+        - podAffinityTerm:
+            labelSelector:
+              matchExpressions:
+                - key: app
```
jawnsy, line 164: I think we should use `app.kubernetes.io/name` instead, the `app` label was a legacy one and the former is a de-facto standard label, as it's listed as recommended https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels

jawnsy (approved):

PR #181, opened 2021-12-10
chore: clean up examples

jawnsy:
* Update OpenShift example to create Route with Ingress object
* Remove superfluous configuration options
* Enable lint check for resource request/limit settings

### DIFF
--- a/examples/ingress/ingress.values.yaml
+++ b/examples/ingress/ingress.values.yaml
@@ -1,26 +1,14 @@
-# Using coder with ingress in versions from 1.21 and newer
-# Coder's built-in ingress controller is no longer packaged
-# the coderd pod does not require a fanout so it is exposed
-# by creating a LoadBalancer service
 coderd:
-  serviceNext: true
   devurlsHost: '*.devurls.coderhost.com'
   serviceSpec:
+    # The Ingress will route traffic to the internal ClusterIP.
     type: ClusterIP
-    # The values.yaml file in the chart includes LoadBalancer
-    # specs which need to have the keys removed using null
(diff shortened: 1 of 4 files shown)


johnstcn (approved):
I haven't tried these changes myself but they look fine from my side!

examples/ingress/ingress.values.yaml
```diff
@@ -1,26 +1,14 @@
...
 coderd:
-  serviceNext: true
   devurlsHost: '*.devurls.coderhost.com'
   serviceSpec:
+    # The Ingress will route traffic to the internal ClusterIP.
     type: ClusterIP
-    # The values.yaml file in the chart includes LoadBalancer
-    # specs which need to have the keys removed using null
-    # this does not work as a sub-chart
-    loadBalancerIP: null
```
mterhar, line 13: @jawnsy I see this is already merged but as long as these loadBalancerIP, externalTrafficPolicy, and loadBalancerSourceRanges values are in the `values.yaml` file https://github.com/cdr/enterprise-helm/blob/main/values.yaml#L19 , the user is required to nullify them when changing the type to ClusterIP or NodePort.  Is the coderd service spec going to be changed to remove these lines?

jawnsy: these seem to just be ignored for ClusterIP services, did you see something different?

mterhar: Kubectl doesn't complain about it but I added it because users were seeing errors about "service type ClusterIP doesn't support loadBalancerIP" and Helm uses `null` to remove the key from the yaml output.

PR #186, opened 2021-12-10
feat: make proxy settings configurable

jawnsy:
* Add coderd proxy settings for outbound connections
* Move utility functions into separate file
* Extract LoadChart out of TestDeployment tests to avoid repeatedly
  loading the chart for parallel tests
* Refactor Render function to accept a function closure, as with
  MustRender, to make the function signatures more consistent
* Add utility function to convert hard-coded environment variables
  into a map


jawnsy:
I tested this using an air-gapped installation. The cluster does not have external network access, and the only path to the Internet is a dual-homed host with a proxy installed.

Using the following values:

```yaml
coderd:
  proxy:
    http: http://10.182.0.2:8888
    exempt: ghcr.io,cluster.local
```

I verified that:

* In-cluster traffic works correctly (coderd can communicate with kubernetes API, which does not use the proxy)
* Importing images from ghcr.io times out, since we try to connect directly and the system is air-gapped
* Importing images from GCR works correctly (since images use the internal Private Google Access network)
* Importing images from Docker Hub works correctly (they can be tracked by Coder, but not actually pulled, since the worker node is not configured to pull through the proxy)

[sc-19357]

johnstcn (approved):

jawnsy:
For posterity, this article surveys the different environment variable options and how they work: https://about.gitlab.com/blog/2021/01/27/we-need-to-talk-no-proxy/

PR #199, opened 2021-12-18
feat: add recommended labels for all objects

jawnsy:
Add the Standard Labels that Helm describes in its best practices
documentation: https://helm.sh/docs/chart_best_practices/labels/

### DIFF
--- a/templates/coderd.yaml
+++ b/templates/coderd.yaml
@@ -2,13 +2,17 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  labels:
-    app.kubernetes.io/name: {{ include "coder.serviceName" . }}
-    app.kubernetes.io/part-of: coder
-    app: {{ include "coder.serviceName" . }}
-    coder.deployment: {{ include "coder.serviceName" . }}
   name: {{ include "coder.serviceName" . }}
   namespace: {{ .Release.Namespace | quote }}
+  labels:
+    app.kubernetes.io/name: {{ .Chart.Name }}
+    helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/version: {{ .Chart.AppVersion }}
+    app.kubernetes.io/component: {{ include "coder.serviceName" . }}
+    app: {{ include "coder.serviceName" . }}
+    coder.deployment: {{ include "coder.serviceName" . }}
   annotations: {{ toYaml .Values.services.annotations | nindent 4 }}
 spec:
   replicas: {{ default 1 .Values.coderd.replicas }}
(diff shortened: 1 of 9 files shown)


templates/coderd.yaml
```diff
@@ -23,7 +27,9 @@ spec:
   template:
```
jawnsy, line 27: unfortunately, matchLabels is immutable, so we can't add the new labels here.

there might be interesting behavior if someone installs two separate releases coder deployments into the same namespace, but I guess that's not possible since the names would conflict anyways

templates/coderd.yaml
```diff
@@ -229,6 +242,13 @@ metadata:
   name: {{ include "coder.serviceName" . }}
   {{- end }}
   namespace: {{ .Release.Namespace | quote }}
+  labels:
+    app.kubernetes.io/name: {{ .Chart.Name }}
```
jawnsy, line 246: these labels make it possible to query for all objects corresponding to a given version of Coder, and let you search (e.g. list all coder 1.25 installs in a cluster)

tests/defaults_test.go
```diff
@@ -50,23 +52,48 @@ func TestNamespace(t *testing.T) {
...
 				metaObject, err := meta.Accessor(obj)
-				require.NoError(t, err, "failed to get object metadata")
+				require.NoErrorf(t, err, "failed to get object metadata for object %q with name %q",
+					obj.GetObjectKind().GroupVersionKind(), metaObject.GetName())
 
+				// Verify that all objects are using the supplied namespace
 				actualNamespace := metaObject.GetNamespace()
-				require.Equal(t, namespace, actualNamespace,
-					"deployed namespace does not match target")
+				require.Equalf(t, namespace, actualNamespace,
```
jawnsy, line 76: This check ensures that all created objects have the expected labels, or else emits an error so we can identify the object missing stuff. I'm not sure if there's a more concise way to note the debug info (object kind and name), which is needed to uniquely identify a given resource

templates/rbac.yaml
```diff
@@ -15,13 +22,29 @@ metadata:
...
   annotations: {{ toYaml .Values.coderd.builtinProviderServiceAccount.annotations | nindent 4 }}
-  labels: {{ toYaml .Values.coderd.builtinProviderServiceAccount.labels | nindent 4 }}
+  labels:
+    app.kubernetes.io/name: {{ .Chart.Name }}
+    helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/version: {{ .Chart.AppVersion }}
+    app.kubernetes.io/component: {{ include "coder.serviceName" . }}
+    {{- with .Values.coderd.builtinProviderServiceAccount.labels -}}
```
jawnsy, line 32: this behavior is similar for extraLabels on deployments, but unfortunately this name can't be changed without breaking backward compatibility

johnstcn (approved):
Thanks for adding these!

PR #205, opened 2022-01-04
fix(helm): allow istio traffic in network policies

johnstcn:
Turns out that istio doesn't like the new network policy rules.
Instead of adding rules to specifically work with every service mesh out there, just allow all egress traffic from coderd and timescale.

### DIFF
--- a/templates/networkpolicies.yaml
+++ b/templates/networkpolicies.yaml
@@ -93,13 +93,16 @@ spec:
   # Allow egress to istio service mesh. This shouldn't cause any issues if
   # Istio is not in use.
   egress:
+  # Allow all traffic to Istio control plane. This may need to be adjusted
+  # based on your cluster configuration.
   - to:
     - namespaceSelector:
         matchLabels:
           app.kubernetes.io/name: istio-controlplane
-    - podSelector:
+      podSelector:
         matchLabels:
           istio: pilot
+  # Allow all outbound DNS traffic (53/UDP and TCP)
   - ports:
     - port: 53
       protocol: TCP
--- a/templates/networkpolicies.yaml
+++ b/templates/networkpolicies.yaml
@@ -43,7 +43,6 @@ spec:
       app.kubernetes.io/component: coderd
   policyTypes:
     - Ingress
-    - Egress
   # Deny all ingress traffic, except on our service ports
   ingress:
     - from: []
@@ -52,9 +51,6 @@ spec:
           port: 8080
         - protocol: TCP
           port: 8443
-  # Explicitly allow all egress traffic
-  egress:
-    - {}
 {{- end }}
 
 {{/* Policies for the built-in PostgreSQL database */}}
@@ -94,7 +90,19 @@ spec:
       ports:
         - protocol: TCP
           port: 5432
-  # Explicitly allow all egress traffic
+  # Allow egress to istio service mesh. This shouldn't cause any issues if
+  # Istio is not in use.
   egress:
-    - {}
+  - to:
+    - namespaceSelector:
+        matchLabels:
+          app.kubernetes.io/name: istio-controlplane
+    - podSelector:
+        matchLabels:
+          istio: pilot
+  - ports:
+    - port: 53
+      protocol: TCP
+    - port: 53
+      protocol: UDP
 {{- end }}
--- a/templates/networkpolicies.yaml
+++ b/templates/networkpolicies.yaml
@@ -43,6 +43,7 @@ spec:
       app.kubernetes.io/component: coderd
   policyTypes:
     - Ingress
+    - Egress
   # Deny all ingress traffic, except on our service ports
   ingress:
     - from: []
@@ -51,6 +52,9 @@ spec:
           port: 8080
         - protocol: TCP
           port: 8443
+  # Explicitly allow all egress traffic
+  egress:
+    - {}
 {{- end }}
 
 {{/* Policies for the built-in PostgreSQL database */}}
@@ -90,6 +94,7 @@ spec:
       ports:
         - protocol: TCP
           port: 5432
-  # Deny all egress traffic
-  egress: []
+  # Explicitly allow all egress traffic
+  egress:
+    - {}
 {{- end }}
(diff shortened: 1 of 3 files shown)


johnstcn:
[sc-20008]

coadler (approved):

johnstcn:
```yaml
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
    [...]
spec:
  egress:
  - {}
  ingress:
  - ports:
    - port: 8080
      protocol: TCP
    - port: 8443
      protocol: TCP
  podSelector:
    matchLabels:
      app.kubernetes.io/component: coderd
      app.kubernetes.io/instance: coder
  policyTypes:
  - Ingress
  - Egress
```

```shell
$ kubectl describe networkpolicies.networking.k8s.io coderd 
Name:         coderd
Namespace:    [...]
Created on:   2022-01-04 17:49:40 +0000 UTC
Labels:       <none>
Annotations:  <none>
Spec:
  PodSelector:     app.kubernetes.io/component=coderd,app.kubernetes.io/instance=coder
  Allowing ingress traffic:
    To Port: 8080/TCP
    To Port: 8443/TCP
    From: <any> (traffic not restricted by source)
  Allowing egress traffic:
    To Port: <any> (traffic allowed to all ports)
    To: <any> (traffic not restricted by destination)
  Policy Types: Ingress, Egress
```

templates/networkpolicies.yaml
```diff
@@ -51,6 +52,9 @@ spec:
           port: 8080
         - protocol: TCP
           port: 8443
+  # Explicitly allow all egress traffic
+  egress:
```
jawnsy, line 56: I think this rule would override anything admins set, since Kubernetes applies the union of all policies. Adding this means that admins have no way of locking down egress from coderd without managing network policies themselves. Is that the behavior we want?

>Network policies do not conflict; they are additive. If any policy or policies apply to a given pod for a given direction, the connections allowed in that direction from that pod is the union of what the applicable policies allow. Thus, order of evaluation does not affect the policy result.

https://kubernetes.io/docs/concepts/services-networking/network-policies/

My understanding was that the default behavior is to allow all egress, especially if we don't specify policyTypes: Egress - the same page says:

>By default, a pod is non-isolated for egress; all outbound connections are allowed. A pod is isolated for egress if there is any NetworkPolicy that both selects the pod and has "Egress" in its policyTypes; we say that such a policy applies to the pod for egress. When a pod is isolated for egress, the only allowed connections from the pod are those allowed by the egress list of some NetworkPolicy that applies to the pod for egress. The effects of those egress lists combine additively.

templates/networkpolicies.yaml
```diff
@@ -90,6 +94,7 @@ spec:
       ports:
         - protocol: TCP
           port: 5432
-  # Deny all egress traffic
-  egress: []
+  # Explicitly allow all egress traffic
+  egress:
```
jawnsy, line 98: I think we should only allow egress to Istio here, since otherwise timescale really has no need to initiate connections? Admins that want to allow all can disable the policies entirely

johnstcn: We can add explicit rules for individual service meshes, but we won't know in advance which ones folks will be using.

jawnsy: Yeah, but the workaround is simple (disable network policies, with the corresponding Helm value) and we can always extend the rules later if needed. I'd prefer to keep things locked down as much as possible

templates/networkpolicies.yaml
johnstcn: > admins have no way of locking down egress from coderd without managing network policies themselves. Is that the behavior we want?

I think that's reasonable behaviour. If admins want to override them, they can disable our network policies and manage network policies themselves.

templates/networkpolicies.yaml
```diff
@@ -90,6 +90,19 @@ spec:
...
-  # Deny all egress traffic
-  egress: []
+  # Allow egress to istio service mesh. This shouldn't cause any issues if
+  # Istio is not in use.
+  egress:
+  - to:
+    - namespaceSelector:
+        matchLabels:
+          app.kubernetes.io/name: istio-controlplane
+    - podSelector:
```
jawnsy, line 100: can these be combined? right now, it's checking for the namespace OR pod label, but I think we want an AND here?

tests/network_policy_test.go
```diff
@@ -79,8 +79,7 @@ func TestNetworkPolicyCoder(t *testing.T) {
 			require.Equal(t, test.ExpectCoderPolicy, exist, "coderd network policy")
 			if test.ExpectCoderPolicy {
 				require.Contains(t, policy.Spec.PolicyTypes, networkingv1.PolicyTypeIngress, "expected to restrict ingress")
-				require.NotContains(t, policy.Spec.PolicyTypes, networkingv1.PolicyTypeEgress, "expected all egress to be allowed")
```
jawnsy, line 82: can we revert this change, since the Coder change was reverted?

templates/networkpolicies.yaml
```diff
@@ -90,6 +90,19 @@ spec:
...
+        matchLabels:
+          istio: pilot
+  - ports:
```
jawnsy, line 103: this should also be part of the "to", not a separate item - this allows outbound port 53 to anything (check the describe output)

tests/network_policy_test.go
```diff
@@ -117,7 +116,6 @@ func TestNetworkPolicyCoder(t *testing.T) {
 			if test.ExpectDatabasePolicy {
 				require.Contains(t, policy.Spec.PolicyTypes, networkingv1.PolicyTypeIngress, "expected to restrict ingress")
 				require.Contains(t, policy.Spec.PolicyTypes, networkingv1.PolicyTypeEgress, "expected to restrict egress")
-				require.Empty(t, policy.Spec.Egress, "expected empty egress rules")
```
jawnsy, line 120: should add an assertion that the policy is as we expect, the Go struct is probably more obvious about how it's working than the YAML

templates/networkpolicies.yaml
jawnsy: This is a really handy tool to visualize how the policies work - unfortunately the policy language is kind of badly designed, which leads to a lot of gotchas :frowning: https://editor.cilium.io/

jawnsy: more info about this behavior here: https://kubernetes.io/docs/concepts/services-networking/network-policies/#behavior-of-to-and-from-selectors

johnstcn: I tested this out and the two podSelectors can be combined, but we need to allow all outbound TCP/UDP traffic for DNS to function.

jawnsy (approved):

PR #209, opened 2022-01-13
feat: add client certs to coderd

mterhar:
This is designed to expose configs that allow admins to set client certs which will be picked up by Coderd. 

Followed the same conventions put forth for the postgres client SSL auth though I didn't see a "root cert" so I'm not sure if that's applicable. 

This has not been tested since I don't have a git provider or docker registry that uses client side authentication.


templates/_common.tpl
```diff
@@ -124,6 +134,16 @@ volumeMounts:
     mountPath: /etc/ssl/certs/pg/rootcert
     readOnly: true
 {{- end }}
+{{- if ne .Values.coderd.clientSsl.certSecret.name "" }}
+  - name: pgcert
```
jawnsy, line 138: pgcert doesn't look correct?

templates/_common.tpl
```diff
@@ -86,6 +86,16 @@ volumes:
     secret:
       secretName: {{ .Values.postgres.ssl.rootCertSecret.name | quote }}
 {{- end }}
+{{- if ne .Values.coderd.clientSsl.certSecret.name "" }}
```
jawnsy, line 89: why ssl instead of tls?

values.yaml
```diff
@@ -54,6 +54,20 @@ coderd:
     # coderd.tls.devurlsHostSecretName -- The secret to use for DevURL TLS.
     devurlsHostSecretName: ""
 
+  clientSsl:
```
jawnsy, line 57: please add some unit tests, @coadler can probably help you

johnstcn (changes requested):
Thanks for this! I don't have any comments besides what Jonathan already highlighted. Once those are addressed, lgtm. 

For testing, you can spin up a local Docker registry and put that behind a nginx container using a config like [this one](https://github.com/idlerun/docker-registry-setup/blob/master/nginx-registry.conf)

templates/_common.tpl
coadler: we should probably use tls here. the postgres certs all reference ssl because that's how they're named in the postgres documentation and settings.

mterhar: I did it to match PG but if that's specific to their docs, I'll make it TLS.

tests/go.mod
```diff
@@ -10,8 +10,8 @@ require (
 	golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1
 	helm.sh/helm/v3 v3.8.0
 	k8s.io/api v0.23.1
-	k8s.io/apimachinery v0.23.1
-	k8s.io/utils v0.0.0-20210930125809-cb0fa318a74b
+	k8s.io/apimachinery v0.23.3
```
jawnsy, line 13: dependency updates like this should ideally be in a separate PR, because it makes it easier to revert independently if needed

mterhar: Will revert this change for now.

tests/values.go
```diff
@@ -94,6 +94,12 @@ type CoderdValues struct {
 	Proxy                         *CoderdProxyValues                         `json:"proxy" yaml:"proxy"`
 	ReverseProxy                  *CoderdReverseProxyValues                  `json:"reverseProxy" yaml:"reverseProxy"`
 	NetworkPolicy                 *CoderdNetworkPolicyValues                 `json:"networkPolicy" yaml:"networkPolicy"`
+	ClientTls                     *CoderdClientTlsValues                     `json:"clientTls" yaml:"clientTls"`
+}
+
```
jawnsy, line 99: needs corresponding tests, @ericpaulsen is working on these for something else and might be able to help?

tests/values.go
```diff
@@ -94,6 +94,12 @@ type CoderdValues struct {
 	Proxy                         *CoderdProxyValues                         `json:"proxy" yaml:"proxy"`
 	ReverseProxy                  *CoderdReverseProxyValues                  `json:"reverseProxy" yaml:"reverseProxy"`
 	NetworkPolicy                 *CoderdNetworkPolicyValues                 `json:"networkPolicy" yaml:"networkPolicy"`
+	ClientTls                     *CoderdClientTlsValues                     `json:"clientTls" yaml:"clientTls"`
```
jawnsy, line 97: I think Go prefers using uppercase for acronyms, e.g. `ClientTLS`

values.yaml
```diff
@@ -51,6 +51,20 @@ coderd:
...
+    certSecret:
+      # coderd.clientTls.certSecret.name -- Name of the secret.
+      name: ""
+      # coderd.clientTls.certSecret.key -- Key pointing to a certificate in the secret.
+      key: ""
+    # coderd.clientTls.keySecret -- Secret containing a PEM encoded key file.
+    keySecret:
+      # coderd.clientTls.keytSecret.name -- Name of the secret.
+      name: ""
+      # .keySecret.key -- Key pointing to a certificate in the secret.
```
jawnsy, line 65: should be coderd.clientTls.keySecret.key

values.yaml
```diff
@@ -51,6 +51,20 @@ coderd:
...
+  clientTls:
+    # coderd.clientTls.certSecret -- Secret containing a PEM encoded cert file.
+    certSecret:
+      # coderd.clientTls.certSecret.name -- Name of the secret.
+      name: ""
+      # coderd.clientTls.certSecret.key -- Key pointing to a certificate in the secret.
+      key: ""
+    # coderd.clientTls.keySecret -- Secret containing a PEM encoded key file.
+    keySecret:
+      # coderd.clientTls.keytSecret.name -- Name of the secret.
```
jawnsy, line 63:
```suggestion
      # coderd.clientTls.keySecret.name -- Name of the secret.
```

examples/clientcerts/clientcerts.values.yaml
```diff
@@ -0,0 +1,28 @@
...
+  ssl:
+    certSecret:
+      name: "database-client-cert-secret"
+      key: "database-client-cert-filename.crt"
+    keySecret:
+      name: "database-client-key-secret"
+      key: "database-client-key-filename.key"
+    rootCertSecret:
+      name: "database-root-cert-secret"
+      key: "database-root-cert-filename.crt"
```
jawnsy, line 28: needs a blank line at end of file

README.md
```diff
@@ -25,11 +25,15 @@ View [our docs](https://coder.com/docs/setup/installation) for detailed installa
...
 | coderd.builtinProviderServiceAccount | object | Customize the built-in Kubernetes provider service account. | `{"annotations":{},"labels":{}}` |
 | coderd.builtinProviderServiceAccount.annotations | object | A KV mapping of annotations. See: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/ | `{}` |
 | coderd.builtinProviderServiceAccount.labels | object | Add labels to the service account used for the built-in provider. | `{}` |
+| coderd.clientTls.certSecret | object | Secret containing a PEM encoded cert file. | `{"key":"","name":""}` |
+| coderd.clientTls.certSecret.key | string | Key pointing to a certificate in the secret. | `""` |
+| coderd.clientTls.certSecret.name | string | Name of the secret. | `""` |
+| coderd.clientTls.keySecret | object | Secret containing a PEM encoded key file. | `{"key":"","name":""}` |
```
jawnsy, line 36: keySecret.key and keySecret.name don't seem documented for some reason?

examples/clientcerts/clientcerts.values.yaml
```diff
@@ -0,0 +1,28 @@
+# These sections are used for the client side of certificate authentication.
+coderd:
+  clientTls:
```
jawnsy, line 3: since there's already a coderd.tls section, maybe this could be coderd.tls.clientCertSecretName, coderd.tls.clientCertSecretKey to match?

jawnsy: coderd.tlsClient.certSecret.name might be reasonable too, if you want to keep them grouped together. Having similar names just helps these come up in code completions

jawnsy: @deansheather and @Emyrk discovered today that we're using TLS secrets for the other stuff, so we could do the same here (as with the way we configure DevURL things)

Emyrk: Yesssss

ericpaulsen:
@jawnsy - just pushed a commit that includes unit testing (written with the help of @deansheather).

tests/client_tls_test.go
```diff
@@ -0,0 +1,83 @@
...
+			assert.Equal(t, c.secretName, v.Secret.SecretName)
+		})
+
+		AssertVolumeMount(t, coderdCtr.VolumeMounts, c.volName, func(t testing.TB, v v1.VolumeMount) {
+			assert.Equal(t, c.mountPath, v.MountPath)
+			assert.True(t, v.ReadOnly)
+		})
+
+		assert.Equal(t, envVars[c.envName], filepath.Join(c.mountPath, c.secretKey))
+	}
```
deansheather, line 82: This would be a good candidate for a helper IMO, we often have volumes that we only mount once and we'd like to be able to assert that the volume and the volume mount have the expected values.

deansheather: I don't expect it to be done in this PR though

ericpaulsen: Where do we write helpers? I'm interested in taking this on.

deansheather: It'd be in utils.go, something like

```go
func AssertVolumeAndMount(t testing.TB, vols []v1.Volume, volMounts []v1.VolumeMount, name string, fn func(t testing.TB, v v1.Volume, vm v1.VolumeMount)) {}
```

examples/clientcerts/clientcerts.values.yaml
```diff
@@ -0,0 +1,28 @@
+# These sections are used for the client side of certificate authentication.
```
deansheather, line 1: need to change this now

values.yaml
```diff
@@ -51,6 +51,11 @@ coderd:
     # coderd.tls.devurlsHostSecretName -- The secret to use for DevURL TLS.
     devurlsHostSecretName: ""
 
+  # coderd.clientTls -- Client-side TLS configuration for coderd.
+  clientTls:
+    # secretName -- Secret containing a PEM encoded cert file.
```
deansheather, line 56:
```suggestion
    # coderd.clientTls.secretName -- Secret containing a PEM encoded cert file.
```

values.yaml
```diff
@@ -51,6 +51,11 @@ coderd:
     # coderd.tls.devurlsHostSecretName -- The secret to use for DevURL TLS.
     devurlsHostSecretName: ""
 
+  # coderd.clientTls -- Client-side TLS configuration for coderd.
+  clientTls:
```
deansheather, line 55: maybe we should change this to clientTLS too to match the Go way

deansheather (approved):

examples/clientcerts/clientcerts.values.yaml
ericpaulsen: Changed this to use a TLS-type secret. So it's simply `coderd.clientTLS.secretName`.

README.md
ericpaulsen: My latest push adds docs.

examples/clientcerts/clientcerts.values.yaml
ericpaulsen: updated examples file with the new values.

values.yaml
ericpaulsen: it's now `coderd.clientTLS.secretName`.

tests/values.go
ericpaulsen: done!

johnstcn (approved):
Looks good, thanks for doing thing @mterhar and @ericpaulsen !

One small thing that always catches me out is that README.md needs to be updated after any change by running

```
make README.md
```

I went ahead and did this.

PR #211, opened 2022-01-25
fix: coderd.serviceAnnotations

ericpaulsen:
removes `coderd.serviceAnnotations` from the values file, as it was removed from the `coderd` template file in #169.

let me know if we should remove line [82 in `/tests/values.go`](https://github.com/coder/enterprise-helm/blob/24a7a3efd3ccb8b8103e0ecaa888ba0de05de12e/tests/values.go#L82).


deansheather (approved):
Looks good, and you should remove it from that Go file too 😄

PR #217, opened 2022-01-28
feat: test coverage for services.annotations

ericpaulsen:
unit test for the `services.annotations` value.


tests/annotations_test.go
```diff
@@ -0,0 +1,45 @@
...
+	expected := map[string]string{
+		"key": "value",
+		"service.beta.kubernetes.io/aws-load-balancer-backend-protocol": "http",
+	}
+
+	objs := chart.MustRender(t, func(cv *CoderValues) {
+		cv.Services.Annotations = expected
+	})
+
+	depl := MustFindDeployment(t, objs, "coderd")
```
jawnsy, line 25: based on the docs, it looks like services.annotations applies the annotations to all services, whereas coderd.serviceAnnotations only applied the annotations to the coderd service. I'm not sure what the desired behavior should be here, actually. Looks like @deansheather added this in https://github.com/coder/enterprise-helm/pull/134 so maybe he can comment. Maybe the correct solution here is to bring back Dean's change, and use coderd.serviceAnnotations to control annotations on coderd only?

As a user, I think that I'd expect `service.annotations` to apply to all services (`timescale`, `coderd`) and `coderd.serviceAnnotations` only to apply to coderd. But it seems the previous behavior didn't apply `service.annotations` to the coderd service.

deansheather: You removed `coderd.serviceAnnotations` in #169

deansheather (approved):

tests/annotations_test.go
jawnsy: yea, that was unintentional though. I'm not sure what the desired behavior should be here, because:

1. the previous behavior (prior to my change) was that services.annotations was used for ?? and did not apply to coderd (I would have expected it'd apply to coderd too, as well as other services we create)
2. the previous behavior (prior to my change) was that coderd.serviceAnnotations created annotations that apply to the coderd service
3. the current behavior (after my change) applies services.annotations to coderd and other services, but of course, now coderd.serviceAnnotations doesn't do anything
   https://github.com/coder/enterprise-helm/blob/8aef31f7a0c135a073bf9be75523e7fad55b3a49/templates/coderd.yaml#L240-L256

   I'm not sure why we're applying the annotations to the StatefulSet, rather than the timescale Service:

   https://github.com/coder/enterprise-helm/blob/0da85a6f4adc20f9e418e8f8da29ce9a978dc7c3/templates/timescale.yaml#L24-L36

   and I don't know why we're applying this to coderd's deployment:

   https://github.com/coder/enterprise-helm/blob/0da85a6f4adc20f9e418e8f8da29ce9a978dc7c3/templates/coderd.yaml#L2-L16

@deansheather so my question for you is whether we need to be able to control service annotations on a service-by-service basis (e.g. the timescale annotations vs the coderd annotations), in which case the current behavior is broken (we can set annotations for all services but not for coderd only)

Maybe the naming was confusing - does "services" refer to k8s services, or is it referring to coderd/timescale collectively as "services"? The code seems to do the latter, in which case the `coderd.serviceAnnotations` thing makes sense

deansheather: I'm unsure but it's better to have annotations for all individual components be separate values IMO. So `coderd.annotations` `coderd.service(\.a|A)nnotations` `timescale.annotations`

ericpaulsen: If we go with such a solution, that would require removing `services.annotations`, thus rendering this test null. I'll hold off merging this until we decide.

tests/annotations_test.go
```diff
@@ -0,0 +1,44 @@
...
+	t.Parallel()
+
+	chart := LoadChart(t)
+
+	expected := map[string]string{}
+
+	objs := chart.MustRender(t, func(cv *CoderValues) {
+		cv.Coderd.Annotations = expected
+		cv.Coderd.ServiceSpec.Annotations = expected
+		cv.Postgres.Default.Annotations = expected
```
ericpaulsen, line 21: @deansheather this accounts for each annotation value being set properly in the Helm chart. I'd be curious to know if there's a better way to set multiple `struct` values to `expected`.

tests/annotations_test.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+// ensures services.annotations values are applied to both coderd deployment &
+// service
+func TestAnnotations(t *testing.T) {
+	t.Parallel()
+
+	chart := LoadChart(t)
+
+	expected := map[string]string{}
```
ericpaulsen, line 16: @jawnsy @deansheather in the prior commit, this read as:

```
expected := map[string]string{
    "key": "value",
    "service.beta.kubernetes.io/aws-load-balancer-protocol": "http",
}
```

I removed the `"key": "value"` requirement because the test was failing. Let me know if this is still needed. Test failure output was:

```
expected: map[string]string{"key":"value"}
actual  : map[string]string{}
```

values.yaml
```diff
@@ -351,12 +356,12 @@ postgres:
...
       enable: true
+    # postgres.default.annotations -- Apply annotations to the default postgres service.
+    # https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/
+    annotations: {}
 
 # services -- Kubernetes Service configuration that applies to Coder services.
 services:
-  # services.annotations -- A KV mapping of annotations. See:
-  # https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/
-  annotations: {}
```
deansheather, line 359: We need to keep this for backwards compatibility and merge with the current values. @jawnsy might know how to merge values together and then toYaml it in the template

tests/annotations_test.go
deansheather, line 16: We definitely should be testing with actual values (and perhaps have a separate test for empty annotations, but I don't think that's as necessary).

I would also recommend testing with three `expected` vars (i.e. `expectedCoderd`, `expectedCoderdService`, `expectedTimescale`) with completely different values so we can make sure the correct values are being used in the correct places.

Also you'll need to add a test for backwards compatibility

deansheather: I replied here https://github.com/coder/enterprise-helm/pull/217#discussion_r799828357

tests/annotations_test.go
deansheather: Not sure what your question is, but this looks OK to me

ericpaulsen: cool - I just wasn't sure if this was the most efficient method in setting multiple values to the `expected` variable.

templates/coderd.yaml
```diff
@@ -13,7 +13,7 @@ metadata:
     app.kubernetes.io/component: {{ include "coder.serviceName" . }}
     app: {{ include "coder.serviceName" . }}
     coder.deployment: {{ include "coder.serviceName" . }}
-  annotations: {{ toYaml .Values.services.annotations | nindent 4 }}
+  annotations: {{ toYaml .Values.coderd.annotations | nindent 4 }}
```
jawnsy, line 16: I think we can use the merge function for this, I haven't tried it but we might be able to do something like:

`{{ toYaml (merge .Values.services.annotations .Values.coderd.annotations) | nindent 4 }}`

https://helm.sh/docs/chart_template_guide/function_list/#merge-mustmerge

deansheather (approved):

ericpaulsen:
@deansheather @jawnsy - I may have got myself into a figurative `git` hairball trying to resolve these conflicts. let me know when you have some time to assist.

README.md
```diff
@@ -25,8 +25,13 @@ View [our docs](https://coder.com/docs/setup/installation) for detailed installa
...
+<<<<<<< HEAD
```
jawnsy, line 28: this file is autogenerated, to fix the conflicts, try running:

```console
rm README.md
make README.md
```

You will need `helm-docs`, which is included in the `coder-dev-ubuntu` image

ericpaulsen: Just pushed a commit doing so, and it still returned conflicts...

templates/timescale.yaml
```diff
@@ -55,7 +55,7 @@ spec:
         app.kubernetes.io/component: {{ include "timescale.serviceName" . }}
         app: timescale
         coder.deployment: timescale
-      annotations: {{ toYaml .Values.services.annotations | nindent 8 }}
+      annotations: {{ toYaml .Values.postgres.default.annotations | nindent 8 }}
```
jawnsy, line 58: looks like this one is missing a merge between services?

templates/coderd.yaml
```diff
@@ -13,7 +13,7 @@ metadata:
     app.kubernetes.io/component: {{ include "coder.serviceName" . }}
     app: {{ include "coder.serviceName" . }}
     coder.deployment: {{ include "coder.serviceName" . }}
-  annotations: {{ toYaml .Values.services.annotations | nindent 4 }}
+  annotations: {{ toYaml (merge .Values.coderd.annotations .Values.services.annotations) | nindent 4 }}
```
jawnsy, line 16: Do this to handle the null case:

```
  annotations: {{ merge (.Values.postgres.default.annotations | default (dict)) (.Values.services.annotations | default (dict)) | quote }}
```

should also add test cases for this, it makes the chart a bit more robust :)

ericpaulsen: awesome - thank you for the help!

tests/annotations_test.go
```diff
@@ -0,0 +1,84 @@
...
+	db := MustFindStatefulSet(t, objs, "timescale")
+	assert.Equal(t, mergeAnnotations(expectedGlobal, expectedTimescale), db.Annotations)
+}
+
+func TestAnnotationsEmpty(t *testing.T) {
+	t.Parallel()
+
+	var (
+		chart = LoadChart(t)
+		objs  = chart.MustRender(t, nil)
```
jawnsy, line 59: you need a MustRender function that explicitly sets the annotation values to nil, you should be able to change the chart back to how you had it before, see a test failure, add back the null checks, and see that it has been resolved

templates/coderd.yaml
```diff
@@ -13,7 +13,7 @@ metadata:
     app.kubernetes.io/component: {{ include "coder.serviceName" . }}
     app: {{ include "coder.serviceName" . }}
     coder.deployment: {{ include "coder.serviceName" . }}
-  annotations: {{ toYaml .Values.services.annotations | nindent 4 }}
+  annotations: {{ toYaml (merge (.Values.coderd.annotations | default (dict)) (.Values.services.annotations | default (dict))) | nindent 8 }}
```
jawnsy, line 16: nindent 8 looks suspicious, please double check your indents

ericpaulsen: I'm going to give `nindent 4` a shot. That was the value prior.

ericpaulsen: looks like that didn't work. do we need to include `nindent` after each annotation being merged? e.g.

```yaml
annotations: {{ toYaml (merge (.Values.postgres.default.annotations | default (dict)) | nindent 4 (.Values.services.annotations | default (dict))) | nindent 4 }}
```

I feel like this is needed since each annotation we be on it's own new line.

jawnsy (approved):
LGTM

PR #221, opened 2022-02-14
chore: move trustedHostnames value

deansheather:


templates/coderd.yaml
```diff
@@ -194,7 +194,7 @@ spec:
               value: "/etc/ssl/certs/client/tls.key"
             {{- end }}
             - name: CODER_TRUSTED_HOSTNAMES
```
jawnsy, line 196: should this be renamed as well?

values.yaml
```diff
@@ -109,14 +109,19 @@ coderd:
...
+  # will go to the configured access URL in the admin settings on the dashboard
+  # (or the satellite's primary URL as configured by Helm).
+  #
+  # Example:
+  #   alternateHostnames:
+  #     - coder.example.com
+  #     - primary.example.com
+  #
+  # This is used for having a unified hostname across all satellites. Read more:
+  # TODO: link to docs:
```
jawnsy, line 123: do you know the URL yet, or is that still pending review?

jawnsy (approved):

values.yaml
deansheather: Not yet, it doesn't matter if the TODO makes it into the release though since people refer to the repo for the values info anyways so I can change it after it's merged and not have to worry about cherry picking it again

templates/coderd.yaml
deansheather: It doesn't need to be since it's not user facing, but if you want me to rename it I can

jawnsy: I think that's preferable to avoid confusion if it's not too much work to change it, but no big deal

jawnsy (approved):

PR #223, opened 2022-02-15
feat: fail when using deprecated values

jawnsy:
When using deprecated values, emit an error since those values no
longer have an effect.

---

Emits errors like the following

```
$ helm install coder . --dry-run --set=ingress.tls.hostSecretName="abc"
Error: INSTALLATION FAILED: execution error at (coder/templates/legacy.tpl:22:4): The 'ingress.tls.hostSecretName' setting was was removed in 1.27; use 'coderd.tls.hostSecretName' instead
```

### DIFF
--- a/templates/legacy.tpl
+++ b/templates/legacy.tpl
@@ -1,0 +1,59 @@
+{{- if .Values.cemanager }}
+{{- fail "The 'cemanager' setting was removed in 1.27; use 'coderd' instead" }}
+{{- end }}
+
+{{- if .Values.devurls }}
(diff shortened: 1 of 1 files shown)


templates/legacy.tpl
```diff
@@ -0,0 +1,59 @@
+{{- if .Values.cemanager }}
+{{- fail "The 'cemanager' setting was removed in 1.27; use 'coderd' instead" }}
```
deansheather, line 2: "was deprecated in 1.21 and removed in 1.27"

deansheather (approved):

PR #224, opened 2022-02-22
feat: permit installation on Kubernetes >= 1.19

jawnsy:
* Relax version constraint to allow installation on Kubernetes
  1.19 or later, with a warning message for unsupported versions
  (prior to 1.21)
* Set chart and app version to 1.28.0

### DIFF
--- a/tests/values.go
+++ b/tests/values.go
@@ -471,7 +471,7 @@ func (c *Chart) RenderNotes(fn func(*CoderValues), options *chartutil.ReleaseOpt
 		return "", fmt.Errorf("renderManifests: %w", err)
 	}
 
-	// As a special case, ignore any .txt files (e.g. NOTES.txt)
+	// Find and return the NOTES.txt file
 	for key, value := range manifests {
 		if filepath.Base(key) == "NOTES.txt" {
 			return value, nil
--- a/tests/values.go
+++ b/tests/values.go
@@ -378,14 +378,9 @@ func (c *Chart) Validate() error {
 	return c.chart.Validate()
 }
 
-// Render creates a copy of the default chart values, runs fn to
-// modify those values, then applies those values to the chart,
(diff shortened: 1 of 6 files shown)


coadler (approved):

tests/values.go
```diff
@@ -436,6 +437,61 @@ func (c *Chart) Render(fn func(*CoderValues), options *chartutil.ReleaseOptions,
...
+	if err != nil {
+		return "", fmt.Errorf("failed to create render values: %w", err)
+	}
+
+	manifests, err := engine.Render(c.chart, vals)
+	if err != nil {
+		return "", fmt.Errorf("failed to render Chart: %w", err)
+	}
+
+	// As a special case, ignore any .txt files (e.g. NOTES.txt)
```
deansheather, line 485: reverse comment

tests/values.go
```diff
@@ -436,6 +437,61 @@ func (c *Chart) Render(fn func(*CoderValues), options *chartutil.ReleaseOptions,
...
+
+	vals, err = chartutil.ToRenderValues(c.chart, vals, opts, capabilities)
+	if err != nil {
+		return "", fmt.Errorf("failed to create render values: %w", err)
+	}
+
+	manifests, err := engine.Render(c.chart, vals)
+	if err != nil {
+		return "", fmt.Errorf("failed to render Chart: %w", err)
+	}
```
deansheather, line 483: This is almost entirely identical to the function above, I would put it into a func called `RenderManifests` or something that just returns `manifests, error` then you can do whatever filtering you want in `Render` and `RenderNotes`

deansheather (approved):

tests/values.go
jawnsy: Great idea!

johnstcn (approved):

PR #225, opened 2022-02-23
feat: add pprof and prometheus ports to container spec

coadler:


deansheather (approved):

templates/coderd.yaml
```diff
@@ -99,6 +99,10 @@ spec:
           ports:
             - name: tcp-{{ include "coder.serviceName" . }}
               containerPort: 8080
+            - name: pprof-{{ include "coder.serviceName" . }}
+              containerPort: 6060
```
jawnsy, line 103: do we need to adjust the network policies?

coadler: i'm not entirely sure. i was going to set up [gke workload metrics](https://cloud.google.com/stackdriver/docs/solutions/gke/managing-metrics) first and see if editing the network policies are required

PR #227, opened 2022-03-03
fix: update ingress template TLS values

ericpaulsen:
changed the ingress template to point to the coderd TLS values, resulting in deprecation of the `ingress.tls.hostSecretName` and `ingress.tls.devurlsHostSecretName` values.

### DIFF
--- a/templates/ingress.yaml
+++ b/templates/ingress.yaml
@@ -49,16 +49,16 @@ spec:
               name: tcp-{{ include "coder.serviceName" . }}
   {{- if .Values.ingress.tls.enable }}
   tls:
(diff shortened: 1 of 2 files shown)


deansheather (approved):

ericpaulsen:
@deansheather should we merge without the `fuzz` test passing? or is that not a good idea

jawnsy:
@ericpaulsen No merges with failing checks. Please investigate the failure and fix it. I'm happy to pair with you to debug this.

I'm going to make it a required check, we shouldn't be merging with failures because it would make the main branch fail

ericpaulsen:
@jawnsy understood - I was just in Discord talking through it with Dean, but I had to drop for a call. Will be hopping back on 30 mins or so

PR #230, opened 2022-03-17
feat: Add support for AWS_STS_REGIONAL_ENDPOINTS

lanzrein:
* Add an optional environment variable inside of coderd deployment called `AWS_STS_REGIONAL_ENDPOINTS` value is `regional`
* Add a value in `values.yaml` : `awsStsRegionalEndpoints.enable` (false by default) which activates the environment variable `AWS_STS_REGIONAL_ENDPOINTS`

I have not yet tested this on a coder deployment, I will remove the WIP once I can confirm it works as intended.


lanzrein:
I have tested this on my local deployment the following : 
- Set to `enable: false` -- environment variable is missing. 
- Set to `enable: true` -- adds the env var
- Revert to `enable: false` -- removes the variable as expected.

lanzrein:
@ericpaulsen do you want me to squash some commits ?

ericpaulsen (approved):
this looks good to me. however, prior to merging, i'd like for our engineers to provide their input, in case i'm missing anything.

thank you for contributing!

deansheather:
I'm thinking we might get more customer value if we instead let customers just specify arbitrary environment variables for coderd. I don't want to dismiss your work, though. This is a great addition. Thoughts? @coadler @f0ssel

deansheather:
Sorry about this taking so long, I'm going try to make a separate PR today that adds support for arbitrary env variables on coderd and I'll ping you for review :)

lanzrein:
> Sorry about this taking so long, I'm going try to make a separate PR today that adds support for arbitrary env variables on coderd and I'll ping you for review :)

Hello, if you are busy at the moment I do not mind implementing this feature. How I see it would require to modify this MR to allow for `env` objects rather than a single `awsStsRegionalEndpoints.enable` boolean flag. 
Cheers.

PR #232, opened 2022-03-18
feat: added ingressClassName

stevek-pro:
- Added .Values.ingress.classname and ingressClassName to ingress template spec
- Please review and merge


deansheather:
I'm going to close this PR since it's contents are included in #233. I'll review that one instead 👍

PR #233, opened 2022-03-18
fix: devurlsHost ref in ingress template

stevek-pro:
Ingress template was not working correctly because there is neither a `.Values.devurls` nor a `.Values.devurls.host`. Correct values is in both cases `.Values.coderd.devurlsHost`.


templates/ingress.yaml
```diff
@@ -14,6 +14,7 @@ metadata:
     app.kubernetes.io/component: {{ include "coder.serviceName" . }}
   annotations: {{ toYaml .Values.ingress.annotations | nindent 4 }}
 spec:
+  ingressClassName: {{ .Values.ingress.classname | default "" | quote }}
```
deansheather, line 17:
```suggestion
  ingressClassName: {{ .Values.ingress.className | default "" | quote }}
```

values.yaml
```diff
@@ -269,6 +269,8 @@ coderd:
 ingress:
   # ingress.enable -- A boolean controlling whether to create an Ingress.
   enable: false
+  # Set ingress.classname -- The Classname of Ingress
+  classname: ""
```
deansheather, line 273:
```suggestion
  # ingress.className -- The ingressClassName to set on the Ingress.
  className: ""
```

deansheather (approved):

coadler (approved):

deansheather:
Thank you for your contribution! @stevek-pro

deansheather:
@lanzrein we plan on doing 1.29.2 tomorrow which will include it, and failing that it will be released in 1.30 RC1 next week

PR #235, opened 2022-03-29
feat: allow customizing coderd liveness and readiness probes

johnstcn:


values.yaml
```diff
@@ -201,6 +201,20 @@ coderd:
     limits:
       cpu: "250m"
       memory: "512Mi"
+  
+  # coderd.liveness -- Configure the liveness check for the coderd service.
+  liveness:
+      initialDelaySeconds: 10
+      failureThreshold: 15
+      periodSeconds: 10
+      timeoutSeconds: 3
```
f0ssel, line 210: I know they are equal in the chart currently, but usually the liveness probe is more forgiving than a readiness probe. Readiness determines if traffic can be routed to it, and can be pretty easily put back in traffic. Liveness will kill the pod if it isn't passing in that timeframe, so you effectively "disable" the readiness probe by having them equal since the pod will die at the same time the traffic is being routed away from it. 

I suggest we increase these, my gut says the liveness should be 2-3x the readiness but it's purely a guess.

f0ssel (approved):

values.yaml
johnstcn: So something like this then?

```suggestion
  liveness:
      initialDelaySeconds: 30
      failureThreshold: 45
      periodSeconds: 30
      timeoutSeconds: 10
```

Emyrk (approved):

PR #236, opened 2022-04-05
feat(helm): allow arbitrary environment vars on coderd

deansheather:
Allows arbitrary environment variables to be added to coderd containers through the helm values.

Closes #230 

cc: @lanzrein


johnstcn (approved):
LGTM, but we should probably add a note about what will happen if you try to override one of our existing environment variables. I'm pretty sure they'll get clobbered and lost, but I didn't test this completely.

deansheather:
I'm just gonna say that it's undefined behavior and you shouldn't do it or bad things will happen

johnstcn (approved):

PR #239, opened 2022-04-13
Add option to disable built-in migrations

f0ssel:


johnstcn:
@f0ssel does this need to be backported?

johnstcn (changes requested):
Looks like some unit tests broke!

f0ssel:
Not sure, I'll have to check with the customer but we can do it after merge either way

johnstcn (approved):

PR #242, opened 2022-04-29
fix: Allow UDP traffic between coderd pods for TURN

kylecarbs:
This fixes TURN for larger-scale deployments which require
high-availability.


johnstcn (approved):
Confirmed that this works:
- scaled coderd to 2
- used `nc -u` (with a static binary) to test UDP connectivity

PR #243, opened 2022-05-10
chore: increase default coderd resources

ericpaulsen:
increasing the requests/limits for coderd, given that just 10 concurrent users maxes out the default `250m` CPU allotment. 

[source](https://codercom.slack.com/archives/CTZHPJW0H/p1644561589335069)


deansheather (approved):
this needs to be marked heavily in the upcoming patch notes

coadler:
Agree with Dean. This could potentially cause `coderd` pods to no longer fit in a nodepool.

ericpaulsen:
Good call.

Is it even worth making this change? If it's going to cause problems like this, then I'd prefer to close this PR and have folks manually scale `coderd`.

PR #252, opened 2022-06-29
Modify RBAC roles to be Openshift/OKD compliant

holzman:
When enumerating rules in a role, providing multiple apiGroups and
resources leads to invalid resource apiGroups combinations. E.g.:

 - apiGroups: ["", "apps"]
   resources: ["pods", "deployments"]
   rules: ["create"]

This allows "create" for pods and apps/deployments, but also invalid
combinations like

  apiVersion: v1
  kind: deployments

Openshift/OKD rejects adding these rules - users cannot add permissions
that they do not already have; and users of course don't have access
granted to create invalid resources like deployments.v1.

Signed-off-by: Burt Holzman <burt@fnal.gov>

### DIFF
--- a/templates/rbac.yaml
+++ b/templates/rbac.yaml
@@ -46,12 +46,24 @@ metadata:
     app.kubernetes.io/version: {{ .Chart.AppVersion }}
     app.kubernetes.io/component: {{ include "coder.serviceName" . }}
 rules:
-  - apiGroups: ["", "apps", "networking.k8s.io"] # "" indicates the core API group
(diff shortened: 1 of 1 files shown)


holzman:
This patch also removes any permissions that aren't permitted by standard openshift/OKD users -- in particular, write-access to events (create/update/patch/delete/deletecollection); a few sets of eyes should verify that that functionality isn't needed.  (If it is, the installation docs need to be updated to create a new SCC with those privileges).

ericpaulsen:
closing in favor of #273. thank you for your contribution! @holzman

PR #255, opened 2022-07-19
add: support for environments SA labels/annotations

ericpaulsen:
adds support for annotating & labeling the (previously named) `environments`  service account. updating it to be called `workspaces`.

let me know if unit tests need to be added.


templates/rbac.yaml
```diff
@@ -6,14 +6,17 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   namespace: {{ .Release.Namespace | quote }} 
-  name: environments
```
kylecarbs, line 9: Do we not use this service account t anywhere? I thought it was mounted for the default workspace provider.

ericpaulsen: @kylecarbs i believe just the `coder` service account is mounted to the built-in provider.

i just tested things, and it looks like every workspace is provisioned it's own respective SA, and this `environments` SA is not mounted as the service account. we may need to make some back-end changes to have workspaces default to using this `environments` SA.

thoughts?

PR #258, opened 2022-08-01
add: coderd imagePullSecrets

ericpaulsen:
a customer requested we add `imagePullSecrets` to the `coderd` Deployment, so they can pull the containers from a private registry.

feedback appreciated.


templates/coderd.yaml
```diff
@@ -40,6 +40,8 @@ spec:
     spec:
       securityContext: {{ toYaml .Values.coderd.podSecurityContext | nindent 8 }}
       restartPolicy: Always
+      imagePullSecrets:
+      - name: {{ toYaml .Values.coderd.imagePullSecret }}
```
deansheather, line 44: Should wrap with an `if` so we don't include this is `coderd.imagePullSecret` is empty.

ericpaulsen:
@deansheather let me know if my syntax is correct.

templates/coderd.yaml
```diff
@@ -40,6 +40,8 @@ spec:
     spec:
       securityContext: {{ toYaml .Values.coderd.podSecurityContext | nindent 8 }}
       restartPolicy: Always
+      imagePullSecrets:
+      - name: {{ if .Values.coderd.imagePullSecret }}
```
deansheather, line 44: I think this

```suggestion
      {{- if .Values.coderd.imagePullSecret }}
      imagePullSecrets:
      - name: {{ .Values.coderd.imagePullSecret | quote }}
      {{- end }}
```

deansheather (approved):

PR #262, opened 2022-08-25
feat: toggle verbose logging

ericpaulsen:
a customer asked if we could filter out the `DEBUG` logging in `coderd`, as they're ingesting the logs into DataDog, and the frequency of which is driving up costs.

creates a `true/false` value for the `VERBOSE` environment variable in the `coderd` container.

closes [this issue](https://github.com/coder/v1/issues/12944).


deansheather (approved):
LGTM if it works

ericpaulsen:
I tested 😄

PR #267, opened 2022-10-17
fix: Add `sessionAffinity` to the coderd Service

kylecarbs:
This should prevent clients from being load balanced accross many coderd instances, which ideally results in fewer workspace connections.

### DIFF
--- a/templates/coderd.yaml
+++ b/templates/coderd.yaml
@@ -291,6 +291,7 @@ spec:
   {{- if .Values.coderd.serviceSpec }}
   {{- toYaml .Values.coderd.serviceSpec | nindent 2 }}
   {{- end }}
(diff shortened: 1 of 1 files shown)


johnstcn (approved):

Emyrk (approved):
This could only hurt if it ends up load balancing on 1 node right?

Like if all the sessions land on the same one.

johnstcn:
> This could only hurt if it ends up load balancing on 1 node right?
> 
> Like if all the sessions land on the same one.

Yeah that's my understanding as well; I don't think this is very likely though.

johnstcn:
Opened https://github.com/coder/v1/pull/13204 to update helm submodule

